PR #2, opened 2015-02-19
Balboa Kafka Service

mhotan-s:
Simple service to ingest Kafka Tenant messages to Cassandra


mhotan-s:
I am not that happy with this implementation.  I followed the documentation example pretty thoroughly.  I feel like spinning of threads to handle individual Kafka Streams makes complete sense but I feel I don't know enough about Kafka worse case scenarios.

balboa-kafka/src/main/scala/com/socrata/balboa/kafka/consumer/Consumer.scala
```diff
@@ -0,0 +1,74 @@
...
+  override def onMessage(message: Array[Byte]): Unit = {
+    // I feel this a lot of work mapping bytes to char, creating a string object
+    // convert it to JSON.
+    val m: JsonMessage = new JsonMessage(bytesToString(message))
+
+    try {
+      ds.persist(m.getEntityId, m.getTimestamp, m.getMetrics)
+    } catch {
+      // TODO How should we failed writes to DS.
+      case e: IOException => ???
```
jkew: Hmm. So I think we want to make sure that the BalboaFastFailCheck is being used here to throttle back on our consumption of messages when there is a failure. Basically the way this works is; when we detect a failure with cassandra (or something else) we set a flag to wait a bit before we try again.

balboa-kafka/src/main/scala/com/socrata/balboa/kafka/consumer/Consumer.scala
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * Created by Michael Hotan on 2/18/15.
+ *
+ * Balboa Kafka Consumer
+ */
+sealed abstract class BalboaConsumer(stream: KafkaStream[Array[Byte], Array[Byte]]) extends Runnable {
+
+  override def run(): Unit = {
+    val it: ConsumerIterator[Array[Byte], Array[Byte]] = stream.iterator()
+    while (it.hasNext())
```
jkew: Yeah. There's a good argument for integrating some messages from BalboaFastFailCheck into this. In the JMS version we actually stop/restart the AMQ listeners; but we may be able to do something simpler here; maybe just stop calling .next until the failure resolves itself. You would need to sleep a bit if you are in a failure mode so you don't just have a busy loop.

balboa-kafka/src/main/scala/com/socrata/balboa/kafka/BalboaConsumerGroup.scala
```diff
@@ -0,0 +1,55 @@
...
+
+  /**
+   * Return the streams for a given Balboa Consumer Group.
+   *
+   * Note: Not pure function. How to make pure?
+   *
+   * @param numThreads Number of threads to spin off.
+   */
+  def getStreams(cg: BalboaConsumerGroup, numThreads: Int): Option[List[KafkaStream[Array[Byte],Array[Byte]]]] = cg
+  match {
```
jkew: nit: indenting

README.md
```diff
@@ -76,7 +76,7 @@ resolvers := Seq(
 ```
 libraryDependencies ++= Seq(
   ...
-  "com.socrata" % "balboa-client" % "0.13.0",
+  "com.socrata" % "balboa-client_2.10" % "0.14.0",
```
jkew: It might be a good idea to start integrating krah's sbt socrata style plugins at a future date. It might require a bit cleanup.

balboa-kafka/src/main/resources/config/config.properties
```diff
@@ -0,0 +1,47 @@
...
+
+# The levels at which to summarize. After you start summarizing, you should not
+# ever change these existing. If you do, some tiers of summarization will contain
+# inconsistent data with the rest. 
+balboa.summaries = fifteen_minute,hourly,daily,monthly,yearly
+
+# Period of serialization to use. After you start summarizing, this should never
+# change or things will break really bad. Some datastores (e.g. SQL) use their
+# own serialization and ignore this value.
+balboa.serializer = protobuf
```
jkew: Hmm. I don't think this is used on the consumer side for cassandra.

README.md
mhotan-s: Ill ping Krah about that in a little bit.

balboa-kafka/src/main/scala/com/socrata/balboa/kafka/Main.scala
```diff
@@ -0,0 +1,133 @@
...
+      val threads: Int = props.getProperty("balboa.kafka.topic.partitions", "1").toInt
+      val topic: String = props.getProperty("balboa.kafka.topic")
+
+      val cg: BalboaConsumerGroup = BalboaConsumerGroup.init(consumerConfig, topic)
+      BalboaConsumerGroup.getStreams(cg, threads) match {
+        case Some(streams) =>
+          // For each available stream create a consumer to ingest the stream.
+          val executor = Executors.newFixedThreadPool(threads)
+          streams.foreach(stream => executor.submit(DSConsumer(stream, DataStoreFactory.get())))
+        case None => Console.println("No Available Stream to print")
```
jkew: This error message isn't clear.

balboa-kafka/src/main/scala/com/socrata/balboa/kafka/BalboaConsumerGroup.scala
mhotan-s: Whoops.  Hmm Intellij does not agree that it needs to be indented.  Silly Jetbrains.

balboa-kafka/src/main/scala/com/socrata/balboa/kafka/BalboaConsumerGroup.scala
```diff
@@ -0,0 +1,55 @@
...
+    case ::(cons,topic) =>
+      var topicCountMap: Map[String, Int] = new HashMap[String, Int]
+      topicCountMap += (topic -> numThreads)
```
jkew: Is there any possibility of this being anything other than ( "test_topic" (or whatever) -> # threads). I'm not sure you need to create a mutable hashmap here. Can you just put:
   cons.createMessageStreams(Map(topic -> numThreads)

balboa-kafka/src/main/scala/com/socrata/balboa/kafka/Main.scala
```diff
@@ -0,0 +1,133 @@
...
+ * on a configurable list of ZooKeeper Host Ports.  This service currently listens for all published Tenant Metrics.
+ *
+ */
+object Main extends App {
+
+  /**
+   * Parse handles command line configuration.
+   */
+  val parser = new scopt.OptionParser[BalboaConfig]("balboa-kafka") {
+    opt[File]('f', "configFile") action { (x,c) => c.copy(configFile = x) } validate {
```
jkew: There has to be a cleaner way to format this. I agree it should all be together in one lump here; I just think we need to be cognizant of line length.

balboa-kafka/src/main/scala/com/socrata/balboa/kafka/consumer/Consumer.scala
mhotan-s: Looking into it.  It looks like ActiveMQReceiver in balboa-jms is utilizing ActiveMQSession rollback functionality.  I see how BalboaFastFailCheck is being utilized in Cassandra11QueryImpl though.  "Should" be straight forward implementation.

balboa-client-kafka/src/main/scala/com/socrata/metrics/components/Logger.scala
```diff
@@ -0,0 +1,14 @@
+package com.socrata.metrics.components
```
jkew: I think this file is outside the scope of the consumer changes.

balboa-kafka/src/main/resources/config/config.properties
```diff
@@ -0,0 +1,47 @@
...
+
+zookeeper.connect = localhost:2181
+zookeeper.session.timeout.ms = 1000000
+group.id = test_group
+
+# Balboa defined Kafka Properties
+# The topic this specific service will be listening for
+balboa.kafka.topic = test_topic
+# The number of partitions
+balboa.kafka.topic.partitions = 8
```
jkew: We set a default partition value of 8 here in the config file; but when we get the property we default it to 1. Why?

balboa-kafka/src/main/scala/com/socrata/balboa/kafka/BalboaConsumerGroup.scala
```diff
@@ -0,0 +1,55 @@
...
+   * @param consumerConfig Consumer configurations
+   * @return Balboa Kafka instance
+   */
+  def init(consumerConfig: ConsumerConfig, topic: String): BalboaConsumerGroup = {
+    (consumerConfig, topic) match {
+      case _=> ::(Consumer.create(consumerConfig), topic)
+    }
+  }
+
+  /**
```
jkew: My tendency with javadocs is to leave out obvious bits of information. For instance; clearly numThreads is the number of threads for this consumer. I don't think we need a @param entry for it. You also leave off the consumer group, which is probably fine... but this is an example where a more complicated parameter is not documented, where as the obvious one is.

balboa-kafka/src/main/scala/com/socrata/balboa/kafka/BalboaConsumerGroup.scala
```diff
@@ -0,0 +1,55 @@
...
+import java.util.concurrent.{Executors, ExecutorService}
+
+import com.socrata.balboa.kafka.consumer.TestConsumer
+import kafka.consumer.{KafkaStream, ConsumerConnector, Consumer, ConsumerConfig}
+
+import scala.collection.Map
+import scala.collection.immutable.HashMap
+
+/**
+ * Created by Michael Hotan, michael.hotan@socrata.com
```
jkew: remove ownership info.

balboa-kafka/src/main/scala/com/socrata/balboa/kafka/BalboaConsumerGroup.scala
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * ADT that manages the connection with zookeeper nodes and consumption of Kafka Messages.  ADT attempts to utilize a
```
jkew: ADT = ?
Also; incomplete description.

balboa-kafka/src/main/scala/com/socrata/balboa/kafka/Main.scala
mhotan-s: Yeah I agree, Currently the way the library I am using is limiting our way of formatting this.  I would love to line break for each rule.

balboa-kafka/src/main/scala/com/socrata/balboa/kafka/BalboaConsumerGroup.scala
mhotan-s: Agreed changed

balboa-kafka/src/main/scala/com/socrata/balboa/kafka/BalboaConsumerGroup.scala
```diff
@@ -0,0 +1,55 @@
...
+ * pure functional
+ *
+ * Note: Most likely won't
+ */
+sealed trait BalboaConsumerGroup
+case class ::(consumer: ConsumerConnector, topic: String) extends BalboaConsumerGroup
+
+/**
+ * `BalboaKafka` Companion object
```
jkew: This is not a companion object to BalboaKafka; and anyways... if they are in the same file you don't need to mention it.

balboa-kafka/src/main/scala/com/socrata/balboa/kafka/Main.scala
mhotan-s: New library or we can just remove the hole thing.  I just used CLI interface for testing.

balboa-kafka/src/main/scala/com/socrata/balboa/kafka/consumer/Consumer.scala
```diff
@@ -0,0 +1,74 @@
+package com.socrata.balboa.kafka.consumer
+
+import java.io.IOException
+
+import com.socrata.balboa.metrics.data.DataStore
+import com.socrata.balboa.metrics.impl.JsonMessage
+import kafka.consumer.{ConsumerIterator, KafkaStream}
+
+/**
+ * Created by Michael Hotan on 2/18/15.
```
jkew: authorship -> nix it

balboa-kafka/src/main/scala/com/socrata/balboa/kafka/BalboaConsumerGroup.scala
```diff
@@ -0,0 +1,55 @@
+package com.socrata.balboa.kafka
+
+import java.util.concurrent.{Executors, ExecutorService}
+
+import com.socrata.balboa.kafka.consumer.TestConsumer
```
jkew: Do we really use the TestConsumer in this file?

balboa-kafka/src/main/scala/com/socrata/balboa/kafka/consumer/Consumer.scala
```diff
@@ -0,0 +1,74 @@
...
+   *
+   * @param bytes The bytes to convert.
+   * @return The String representation of the Byte Array.
+   */
+  def bytesToString(bytes: Array[Byte]): String = {
+    new String(bytes.map(_.toChar))
+  }
+
+}
+case class TestConsumer(stream: KafkaStream[Array[Byte], Array[Byte]]) extends BalboaConsumer(stream) {
```
jkew: So I don't see an actual integration test that uses the test consumer yet. It seems like we want to start building the unit tests right away.

jkew:
So beyond nits and documentation I think we have just a couple major issues:
  - no tests
  - not integrated with the fast-fail watcher

Beyond this; as a followup I think we should
  - integrate sbt style checks tools
  - integrate code coverage tools

balboa-kafka/src/main/scala/com/socrata/balboa/kafka/Main.scala
mhotan-s: Agreed misleading error message.  It should be pumped into a log as well.  The documentation for Kafka's "createMessageStreams" is a little ambiguous to what a None optional result means.  I feel like because they used Option it indicates a non error but no streams available for a given topic.   But not sure :-(

balboa-kafka/src/main/resources/config/config.properties
mhotan-s: That was under the assumption that developers (us) who are looking to tweak configuration in the file would have pretty clear understanding of how many threads they need.  1 was purely arbitrary, I just wanted a number that wouldn't confuse people.  But now that I think about it having multiple values for the same property is more confusing.  Changing.

balboa-kafka/src/main/scala/com/socrata/balboa/kafka/BalboaConsumerGroup.scala
mhotan-s: I used it for sanity checks earlier.  Would removing it be preferable?

balboa-kafka/src/main/scala/com/socrata/balboa/kafka/consumer/Consumer.scala
mhotan-s: Right that is something I am aware of.  I was initially a little unsure about the approach so I withheld from test.  But I will take care of that now.

mhotan-s:
Add Signal Handler for graceful shutdown.

PR #4, opened 2015-03-16
Balboa Kafka Client and Kafka Service updates

mhotan-s:
- Refactor of core client libraries.
- Updated SBT and Scala versions.
- Created MetricLoggerToKafka and related components.  This included a slight modification to the core library.  Specifically had MetricLoggerComponent's MetricLogger producer method have a declared returned type of MetricLoggerLike instead of MetricLogger.
- Created KafkaTestHarness and implemented Balboa Kafka tests.  Its the current best we can do with out full dynamic cloud environment creation.  

TODO: Balboa is getting large enough to where we need to refactor the documentation to clearly identify the Core, Common, HTTP, Client, Service structure.


balboa-client-kafka/src/main/scala/com/socrata/metrics/producer/BalboaKafkaProducer.scala
```diff
@@ -0,0 +1,94 @@
...
+import kafka.serializer.Encoder
+
+/**
+ * A wrapper around Kafka [[Producer]].  This Producer is restricted to exactly one Kafka Topic.  The producer is
+ * currently this producer is configured to send messages Syncronously (See: [[Constants.producerDefaultProps]]).
+ *
+ * @tparam K Kafka Key type.  Can be anything as long as the Encoder is appropiately configured
+ * @tparam M Kafka Message type.  Can be anything as long as the Encoder is appropiately configured
+ */
+trait BalboaKafkaProducer[K,M] {
```
splittingfield: Should probable extend Closeable.

balboa-client-kafka/src/main/scala/com/socrata/metrics/producer/BalboaKafkaProducer.scala
```diff
@@ -0,0 +1,94 @@
...
+
+  val producer: Producer[K,M]
```
splittingfield: Should be def's

balboa-client-kafka/src/main/scala/com/socrata/metrics/producer/BalboaKafkaProducer.scala
```diff
@@ -0,0 +1,94 @@
...
+   * Closes the Kafka Producer.
+   * <br>
+   *  See: [[Producer.close]]
+   */
+  def close() = producer.close()
+
+}
+
+// Only One Constructor
+sealed case class ::[K,M](topic: String, producer: Producer[K,M]) extends BalboaKafkaProducer[K,M]
```
splittingfield: Probably not the name you want

balboa-client-kafka/src/main/scala/com/socrata/metrics/producer/BalboaKafkaProducer.scala
splittingfield: Maybe just make it a class?

balboa-client-kafka/src/main/scala/com/socrata/metrics/producer/BalboaKafkaProducer.scala
```diff
@@ -0,0 +1,94 @@
...
+   * @param brokers A list of Kafka brokers.
+   * @param properties Properties to use that overwrite defaults found [[Constants.producerDefaultProps]].  It does not
+   *                   overwrite the key or message encoders.  Those will always be defined in the type parameter.
+   * @tparam K Key type
+   * @tparam M Message type
+   * @tparam KE Encoder type that will be used to encode kafka message "keys"
+   * @tparam ME Encoder type that will be used to encode kafka "messages"
+   * @return Left(error), or Right(KafkaProducer)
+   */
+  def cons[K,M, KE <: Encoder[K]: Manifest, ME <: Encoder[M]: Manifest](topic: String,
```
splittingfield: Probably should be an apply method here

balboa-client-kafka/src/main/scala/com/socrata/metrics/producer/BalboaKafkaProducer.scala
mhotan-s: I just made that into a base class.  producers are now evaluated/constructed at wrapper construction time.

mhotan-s:
I set up a testing harness that spins up a Kafka environment locally for "almost" integration tests.  This Harness actually starts Zookeepers, Kafka Brokers, Consumers and Producers as needed.  Currently, test work and all pass in Intellij but SBT is set up in a fashion where test harness's Zookeeper nodes collide with eachother.  I think zookeepers are failing to be teared down when running in SBT.  One test passes and the rest throw Address already in use exception.

balboa-client-kafka/README.md
```diff
@@ -0,0 +1,92 @@
...
+   "com.socrata" % "balboa-kafka-client" %% "0.14.0",
+   ...
+ )
+```
+
+Others: TODO
+
+## Usage and Code Example
+
+We are currently not allowing the auto creation of topic outside of testing environments to prevent unwanted namespace
```
jkew: Let's move to auto create as discussed, it will make integration significantly easier for other teams, and I don't consider namespace conflicts to be a huge concern (and it would not solve that problem anyways)

balboa-client-kafka/src/main/java/com/socrata/metrics/impl/MetricLoggerJava.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Java wrapper around MetricLoggerToKafka.
+ */
+public class MetricLoggerJava extends AbstractJavaMetricQueue {
+
+    /**
+     * Mapping topics => MetricLoggers.  Allows for the
```
jkew: Sentence ends prematu

balboa-client-kafka/src/main/java/com/socrata/metrics/impl/MetricLoggerJava.java
```diff
@@ -0,0 +1,62 @@
...
+     * Returns a Metric Logger Instance for a given topic.
+     *
+     * @param topic Topic to get Metric Logger for.
+     * @return The producer for a given topic.
+     */
+    public static MetricLoggerJava getInstance(String topic) {
+        if (instances.containsKey(topic)) {
+            return instances.get(topic);
+        } else {
+            MetricLoggerJava logger = new MetricLoggerJava(topic);
```
jkew: this should be in a synchronized block; with a check just before the creation to see if the topic was already added

balboa-client-kafka/src/main/scala/com/socrata/metrics/impl/MetricLoggerToKafka.scala
```diff
@@ -0,0 +1,100 @@
...
+
+  /**
+   * @return The name of the topic that is being used.
+   */
+  def topic: String
+
+  /**
+   * @return The back up file where failed writes will go.
+   */
+  def backupFile: String
```
jkew: I think I would rename this to better reflect that this is for emergency backups. "backupFile" implies that this is a normal condition; but I would go with "emergencyBackupFile" or something to make it clear.

balboa-client-kafka/src/main/scala/com/socrata/metrics/impl/BalboaKafkaComponent.scala
```diff
@@ -0,0 +1,81 @@
...
+        case p: BalboaKafkaProducer =>
+        case _ =>
+          start()
+      }
+
+      producer.send(msg)
+    } catch {
+      case e: FailedToSendMessageException =>
+        Log.error(s"Unable to send message $msg.  writing to emergency file", e)
+        emergencyFileWriter.send(msg)
```
jkew: This is the sort of thing that should also write a message to stderr, not just the log file. I see three possible use cases for the emergencyFile from the producer standpoint
1. We want the producer to keep running if it cannot write messages; but have a fall-back plan
2. We want to lose metrics/messages (if emergencyFile is null)
3. We want to violently shutdown the service if we cannot write metrics

I think the only scenarios we care about right now from the metrics standpoint is 1 or 2.

balboa-client-kafka/src/main/scala/com/socrata/metrics/producer/GenericKafkaProducer.scala
```diff
@@ -0,0 +1,71 @@
...
+    properties match {
+      case Some(other) => p.putAll(other)
+      case None => // NOOP
+    }
+    p.setProperty("serializer.class", manifest[ME].runtimeClass.getName)
+    p.setProperty("key.serializer.class", manifest[KE].runtimeClass.getName)
+
+    brokers match {
+      case Nil =>
+      // TODO Implement Consul Service discovery.
```
jkew: I don't think this would be part of the client library; but part of chef

balboa-client-kafka/src/main/scala/com/socrata/metrics/producer/GenericKafkaProducer.scala
```diff
@@ -0,0 +1,71 @@
...
+ * @param topic Kafka topic to produce to.
+ * @param brokers Kafka Broker metadata list.
+ * @param properties Producer properties.
+ * @tparam K Kafka "Key" type
+ * @tparam M Kafka "Message" type
+ * @tparam KE Encoder for K. Must have a constructor argument signature of `properties: VerifiableProperties = null`
+ * @tparam ME Encoder for M. Must have a constructor argument signature of `properties: VerifiableProperties = null`
+ */
+class GenericKafkaProducer[K,M,KE <: Encoder[K]: Manifest,ME <: Encoder[M]: Manifest](topic: String,
+                                                                                      brokers: List[AddressAndPort] = List.empty,
```
jkew: producers take lists of brokers; consumers connect to zk. I would choose one way or other.

balboa-client-kafka/src/test/scala/com/socrata/integration/kafka/util/BalboaClientTestHarness.scala
```diff
@@ -0,0 +1,178 @@
...
+   * @return Producer
+   */
+  protected def genProducer(topic: String,
+                            brokers: List[AddressAndPort] = List.empty,
+                            properties: Option[Properties] = None): GenericKafkaProducer[K,M,KE,ME]
+
+  /**
+   * This number is assigned by the current number of Kafka servers inside our cluster.
+   */
+  val DEFAULT_SERVER_COUNT = 4
```
jkew: I think this might be unused.

balboa-client-kafka/src/main/scala/com/socrata/metrics/producer/GenericKafkaProducer.scala
jkew: Now I understand why this is the way it is. It's still dumb.

balboa-client-kafka/src/test/scala/com/socrata/integration/kafka/util/BalboaClientTestHarness.scala
```diff
@@ -0,0 +1,178 @@
...
+  /**
+   * See [[BalboaClientTestHarness.genProducer()]].
+   */
+  override def genProducer(topic: String,
+                           brokers: List[AddressAndPort],
+                           properties: Option[Properties]): GenericKafkaProducer[String, String, StringCodec, StringCodec] =
+    StringKafkaProducer(topic, brokers, properties)
+}
+
+trait BalboaMessageClientTestHarness extends BalboaClientTestHarness[String, Message, StringCodec, BalboaMessageCodec] {
```
jkew: love these tests

jkew:
LGTM % comments

balboa-client-kafka/README.md
mhotan-s: This change is completed in socrata-kafka-cookbook version 0.1.20

balboa-client-kafka/src/main/scala/com/socrata/metrics/impl/BalboaKafkaComponent.scala
mhotan-s: Regarding 2, I am not a fan of not having a file to write to.  But I guess I understand that some clients might not care to save messages in case of failure.

balboa-client-kafka/src/main/scala/com/socrata/metrics/producer/GenericKafkaProducer.scala
mhotan-s: What about Dockerized services?

balboa-client-kafka/src/main/scala/com/socrata/metrics/producer/GenericKafkaProducer.scala
mhotan-s: Yeah silly Kafka.  This is kind of why I thought having a consul client embedded in the library would make sense if users chose to use it or not.  It is not terribly crazy because the client would essentially be curling against the consul agent at localhost.

PR #18, opened 2015-06-11
Dockerize Balboa-HTTP

mhotan:


balboa-http/docker/ship.d/run
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh
+
+set -ev
+
+# Parse Jinja2 templates and replace them with configurations.
+/bin/env_parse ${BALBOA_CONFIG}.j2
+/bin/env_parse ${BALBOA_LOG4J}.j2
+
+# Run It!
+exec /usr/bin/java \
```
znep: Make sure to su to the socrata user so we aren't running services as root.

jasonij: An example: https://github.com/socrata/cetera/blob/master/ship.d/run

PR #19, opened 2015-06-15
Balboa Agent Debian Package

mhotan:
Using SBT-Native-Packager, Balboa Agent assembly jars can now be packaged within .deb.  This allows us to publish artifiacts to S3 and utilize existing Socrata Deployment pipeline.


housepage:
:shipit:

PR #24, opened 2015-06-23
Dockerize balboa-service-jms

housepage:


balboa-service-jms/docker/ship.d/run
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
+
+set -ev
+
+# Parse Jinja2 templates and replace them with configurations.
+/bin/env_parse ${BALBOA_CONFIG}.j2
+/bin/env_parse ${BALBOA_LOG4J}.j2
+
+exec sudo -u socrata java \
+  -Dcom.sun.management.jmxremote.port=40507 \
```
znep: You will need to also set com.sun.management.jmxremote.rmi.port (to the same port) to actually be able to talk to it.

mhotan: @housepage I think we want to set that port to $JMX_PORT environment variable that was exposed in the Dockerfile.

PR #23, opened 2015-06-23
Fix example configuration

coyotebush:
The double-percent belongs first. Bump versions.

PR #26, opened 2015-06-25
Fixed mixing project and directory names. 

mhotan:
Having to think of sub project naming is code smell.


balboa-client-dispatcher/README.md
```diff
@@ -27,7 +27,7 @@ SBT:
  ```
  libraryDependencies ++= Seq(
    ...
-   "com.socrata" % "balboa-dispatcher-client" %% "[0.15, )",
+   "com.socrata" % "balboa-dispatcher-client" %% "[0.16, )",
```
coyotebush: Will conflict with #23 but that's okay. Why different version specs in the two READMEs?

mhotan: No good reason but we should settle on a convention.  Must consult SBT docs.

coyotebush: And then let's pick one PR or the other that will change the dependency lines to the form
`"com.socrata" %% "balboa-client-dispatcher" % "0.16.+",`
(updated names, versions, and syntax)

mhotan: Lets update #23 to update all the README to use the + notation for those who are not intuned with mathematical notation.

mhotan:
@housepage Can you give this the final blessing so we can begin to clean up balboa (Incrementally).

housepage:
:two_men_holding_hands:

PR #31, opened 2015-07-06
Removed package dependency on openjdk-headless-jre.

mhotan:
This introduces to string of a dependency.  Debian also does not manage java dependencies very well either.  Therefore we decided to move the java 7 dependency into the debian package recommendation list.


housepage:
:two_men_holding_hands:

PR #32, opened 2015-07-06
Change to use native debian packaging

housepage:


mhotan:
We got this working with JDeb and plan to stick with it.

PR #42, opened 2015-07-21
Retrieve metrics in order by date

### LINKED ISSUE #38: Absolute metrics merged incorrectly
When `Metric.combine(other)` is invoked on an absolute metric, it assumes that the value of `other` should replace the value of `this`. However, `DataStore.find` does not guarantee that its results are ordered by date. As a result, `Metrics.summarize` will merge metrics in a potentially arbitrary order, and the resulting value for an absolute metric will not necessarily be the most recent one.

Possible solutions include:

1. Alter the contract (and implementations) of `DataStore.find` to return results in order by date
2. Store the date in `Metrics` or in each `Metric`, and use this in `Absolution` to decide priority
3. Handle absolute metrics differently, so as to only retrieve the most recent value

coyotebush:
This looks like the most expedient way to fix #38, but it still needs validation.

### DIFF
--- a/balboa-core/src/test/scala/com/socrata/balboa/metrics/data/impl/Cassandra11DataStoreTest.scala
+++ b/balboa-core/src/test/scala/com/socrata/balboa/metrics/data/impl/Cassandra11DataStoreTest.scala
@@ -105,13 +105,13 @@ class Cassandra11DataStoreTest {
       Assert.assertEquals(mock.metricsToReturn, itr.next())
     Assert.assertEquals(List(new AFetch("foo-2498400000", Period.HOURLY),
                              new AFetch("foo-2502000000", Period.HOURLY),
-                             new AFetch("foo-5270400000", Period.HOURLY),
-                             new AFetch("foo-5274000000", Period.HOURLY),
                              new AFetch("foo-2505600000", Period.DAILY),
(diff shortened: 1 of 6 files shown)


balboa-core/src/test/scala/com/socrata/balboa/metrics/data/impl/Cassandra11DataStoreTest.scala
```diff
@@ -209,13 +209,13 @@ class Cassandra11DataStoreTest {
...
     Assert.assertEquals(List(new AFetch("foo-2498400000", Period.HOURLY),
       new AFetch("foo-2502000000", Period.HOURLY),
-      new AFetch("foo-5270400000", Period.HOURLY),
-      new AFetch("foo-5274000000", Period.HOURLY),
       new AFetch("foo-2505600000", Period.DAILY),
       new AFetch("foo-2592000000", Period.DAILY),
+      new AFetch("foo-2678400000", Period.MONTHLY),
       new AFetch("foo-5097600000", Period.DAILY),
       new AFetch("foo-5184000000", Period.DAILY),
-      new AFetch("foo-2678400000", Period.MONTHLY)).sorted,  mock.fetches.sorted)
```
mhotan: For shame

mhotan:
Makes sense :two_men_holding_hands:

PR #45, opened 2015-08-04
Use Travis container infrastructure with caching

coyotebush:
Travis appears to be [encouraging use of their new container-based infrastructure](http://docs.travis-ci.com/user/migrating-from-legacy/), which also allows caching across builds.

I was hoping to see build time improvements, but unfortunately there wasn't any significant change, even on a [rebuild using the cache](https://travis-ci.org/socrata-platform/balboa/builds/73297994).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,3 +4,8 @@ scala:
 jdk:
   - oraclejdk7
   - openjdk7
+sudo: false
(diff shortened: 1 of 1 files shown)


jacobsi:
We no longer use Travis.

PR #54, opened 2015-10-23
Balboa Agent Metrics Update

mhotan:
# Background 
Metrics created in the Metrics Consumer were not being registered with the JMX Consumer.

# Goal
All Metrics registered in the metrics consumer are now registered in the Balboa Agent Registry as well as exposed to the JMX Reporter.

* MetricConsumer Metrics are now registered with Balboa Agent Application
* Single MetricConsumer per Balboa Agent Application.


jacobsi:
This work was completed in a different way in a different PR.

PR #66, opened 2016-02-01
Balboa Agent Refactored

mhotan:
Traded off a time based buffer flushing mechanism with a size based one.

## Purpose
  * Remove ConcurrentModificationException
  * Make MetricConsumer more testable.

## Tests
  * Added 9 unit tests to MetricConsumerSpec.scala

## QA
  * Tested against Ticker-Tape in a live "Staging" environment

## Details
* Balboa Agent now uses Non Singleton Class directly
* Added a configuration attribute for buffersize and sleeptime
* Moved BalboaAgentMetrics into metrics package.
* Renamed some Kafka Staging and RC tests to Scenario Tests
* Refactored Buffer and Item in to MetricsBuffer and MetricsBag
* Included first MetricConsumer tests.


balboa-agent/src/main/java/com/socrata/balboa/agent/MetricConsumer.java
```diff
@@ -106,7 +117,6 @@ private int processNamespace(File dir) {
                 // Flush the file to JMS after every read
 
                 // TODO I don't think we need this so commenting out the below line.
```
jacobsi: Seems like you can get ride of this comment if you delete the line.

balboa-agent/src/main/java/com/socrata/balboa/agent/MetricConsumer.java
```diff
@@ -147,7 +157,6 @@ public Record(String entityId, String name, Number value, long timestamp, Metric
             while (true) {
                 Map<String, String> record = grovel(stream);
                 if (record == null) {
-                    log.error("Unable to process records from {}", filePath);
                     break;
```
jacobsi: What does it mean when a null is returned from grovel? It's not (potentially) an error reading the file?

balboa-agent/src/main/scala/com/socrata/balboa/agent/BalboaAgent.scala
```diff
@@ -52,15 +40,21 @@ object BalboaAgent extends App with Config with LazyLogging {
   logger info s"Starting Balboa Agent.  Consuming metrics from ${dataDir.getAbsolutePath}.  " +
     s"AMQ Server: $activemqServer, AMQ Queue: $activemqQueue $additionalInfo".trim
 
-  val mc = (activemqUser, activemqPassword) match {
-    case (Some(user), Some(password)) => new MetricConsumer(
-      dataDir, MetricJmsQueue.getInstance(user, password, activemqServer, activemqQueue))
-    case _ => new MetricConsumer(dataDir, MetricJmsQueue.getInstance(activemqServer, activemqQueue))
+  val amqConnection = (activemqUser, activemqPassword) match {
+    case (Some(u), Some(p)) => ActiveMQConnection.makeConnection(u, p, activemqServer)
```
jacobsi: Wish you wouldn't do this. 'user' and 'password' are not long variable names, and it makes it really easy to read this line.

balboa-agent/src/main/scala/com/socrata/balboa/agent/BalboaAgent.scala
```diff
@@ -52,15 +40,21 @@ object BalboaAgent extends App with Config with LazyLogging {
...
+  logger.debug(s"Activemq Connection $amqConnection")
+  val mc = new MetricConsumer(dataDir, new MetricJmsQueueNotSingleton(amqConnection, activemqQueue))
+
+  logger.debug(s"Metric Consumer $mc")
+  // TODO (Pre req: Java 8) Use Java Duration class.
+  val future: ScheduledFuture[_] = scheduler.scheduleWithFixedDelay(mc, initialDelay(), interval(), TimeUnit.MILLISECONDS)
 
   Runtime.getRuntime.addShutdownHook(new Thread() {
     override def run(): Unit = {
+      logger info s"Shutting down Balboa Agent.  Attempt to cancel and future subsequent runs."
```
jacobsi: Don't think this has to be a Scala string.

If you put an 'attempt' message in the log, there should be corresponding success and failure messages at the right place. Otherwise, as a log reader, I'm left wondering if something crashed while 'attempting'.

balboa-agent/src/main/scala/com/socrata/balboa/agent/Config.scala
```diff
@@ -11,6 +11,8 @@ object Keys {
 
   lazy val SLEEP_MS = "balboa.agent.sleeptime"
 
+  lazy val INITIAL_DELAY = "balboa.agent.initialdelay"
+
```
jacobsi: Why make a String constant lazy? Is this just continuing the existing pattern?

balboa-agent/src/main/scala/com/socrata/balboa/agent/metrics/BalboaAgentMetrics.scala
```diff
@@ -15,6 +15,10 @@ import scala.util.{Failure, Success, Try}
   */
 object BalboaAgentMetrics {
 
+  /**
+    * NOTE: This also serves as an example of how to
```
jacobsi: How to what?

balboa-agent/src/main/java/com/socrata/balboa/agent/MetricConsumer.java
mhotan: Good question, Looking at it again `grovel` returns null when it reaches the end of the file.

Note that I added this log statement with a previous commit and I deleted it now.  I was under the impression that it meant we have reached the end of a file and this was an acceptable return value from ByteArrayOutputStream.read.  But looking at it now I see that this is infact an error case and could lead to the loss of metrics.  It is strange that there was no previous log message there before.

mhotan: I am going to add a log message back in.

balboa-agent/src/main/scala/com/socrata/balboa/agent/Config.scala
mhotan: Yes.  it really provides no benefit because with the existing way Balboa Agent is instantiated and run it will always require this key.

balboa-client-jms/src/main/java/com/blist/metrics/impl/queue/MetricJmsQueue.java
```diff
@@ -64,7 +64,15 @@ private MetricJmsQueue(ConnectionFactory factory, String queueName) throws JMSEx
         this.factory = factory; // badness happens if it's GC'd
         Connection connection = factory.createConnection();
         realInstance = new MetricJmsQueueNotSingleton(connection, queueName);
-        realInstance.start();
+    }
+
+    @Override
+    public void close() throws Exception {
+        // TODO When a singleton object has two underlying implementations... that is a bad smell.
```
jacobsi: Random TODOs like this are code litter. They so rarely get addressed.

What is bad? A singleton can wrap many underlying objects, that's pretty normal.

The weird thing is naming a class MetricJmsQueueNotSingleton. It's also weird that this instance method closes the static class member. That's definitely weird. Is it because the Java compiler won't let you have a static close method since there is a member method?

balboa-client-jms/src/main/java/com/blist/metrics/impl/queue/MetricJmsQueue.java
```diff
@@ -64,7 +64,15 @@ private MetricJmsQueue(ConnectionFactory factory, String queueName) throws JMSEx
         this.factory = factory; // badness happens if it's GC'd
         Connection connection = factory.createConnection();
         realInstance = new MetricJmsQueueNotSingleton(connection, queueName);
-        realInstance.start();
+    }
+
+    @Override
+    public void close() throws Exception {
```
jacobsi: No super.close?

balboa-client-jms/src/main/java/com/blist/metrics/impl/queue/MetricJmsQueueNotSingleton.java
```diff
@@ -173,7 +69,19 @@ private void create(String entityId, String name, Number value, long timestamp,
...
+        // Flush the buffer if it becomes to large.
+        // A simpler model given the Java default concurrency model.
+        writeBuffer.add(entityId, metrics, timestamp);
+        if (writeBuffer.size() >= JavaJMSClientConfig.bufferSize()) {
+            flushWriteBuffer();
+        }
+    }
+
+    @Override
+    public void close() throws Exception {
```
jacobsi: super.close()?

balboa-agent/src/main/java/com/socrata/balboa/agent/MetricConsumer.java
mhotan: This actually surfaces an issue with the PR.  Because we removed the explicit call to flush the write buffer when the Metric Consumer process is done.  it was a little mysterious because that call was kind of thrown in that method.  I am deleting the comments but need to make a change to make sure all the elements get flushed from the internal buffer without exposing internal representation.  Stand by.

balboa-client/src/main/java/com/blist/metrics/impl/queue/MetricsBuffer.java
```diff
@@ -0,0 +1,89 @@
...
+public class MetricsBuffer {
+    private static final Logger log = LoggerFactory.getLogger(MetricsBuffer.class);
+
+    /**
+     * Class initially copied from Metric Consumer in Coreserver to MetricJmsQueueNotSingleton to here.
+     *
+     * This class was created in an effort to prevent the overloading of a Queueing system.  For legacy
+     * reasons we continue to support it.   This class a couple of known risks that we look to mitigate in
+     * the future.
+     * * Class mutates metrics data in memory.
```
jacobsi: how? why? The why is especially important for anyone who arrives here with the intention of fixing something. Why this solution was chosen gives them reasons to apply to their own solution.

balboa-client/src/main/java/com/blist/metrics/impl/queue/MetricsBuffer.java
```diff
@@ -0,0 +1,89 @@
...
+     * * Additional buffering layer that in memory and can result to non recoverable metric loss.
```
jacobsi: Do you mean if metrics are read from disk, buffered but not transmitted and this process crashes unexpectedly?

Is there a related situation where metrics are read from disk, transmitted, process crashes and the metrics are later re-read and transmitted? Are metrics marked as consumed by deleting the metrics files?

balboa-client/src/main/java/com/blist/metrics/impl/queue/MetricsBuffer.java
```diff
@@ -0,0 +1,89 @@
...
+            internalBuffer.put(bufferKey, notBuffered);
+        }
+    }
+
+    /**
+     * Returns and deletes all existing metrics.
+     *
+     * @return All the current internal metrics.
+     */
+    synchronized Collection<MetricsBag> popAll() {
```
jacobsi: List would be better. You give the API consumer access to more APIs for manipulating the collection, and you explicitly tell them that it is not a set (which implies duplicates are permitted). As long as you don't advertise an order in the API, no specific order can be assumed.

balboa-client/src/test/scala/com/blist/metrics/impl/queue/MetricsBufferSpec.scala
```diff
@@ -0,0 +1,51 @@
...
+  override protected def afterEach(): Unit = {
+    super.afterEach()
+  }
+
+  "A MetricsBuffer" should {
+
+    "Not contain any metrics when none are added" in new EmptyMetrics {
+      assert(metricsBuffer.popAll().size() == 0, "Initial Metrics Buffer strangely contains metrics data.")
+    }
+    "be immutable" in new EmptyMetrics {
```
jacobsi: A MetricsBuffer is not immutable, that is not it's design, nor is that what this test is about.

I can't think of anything very succinct, but maybe something like:

"be isolated from .popAll() results"

balboa-client/src/test/scala/com/blist/metrics/impl/queue/MetricsBufferSpec.scala
```diff
@@ -0,0 +1,51 @@
...
+      // This shows that
```
jacobsi: Probably don't need this.

balboa-client/src/test/scala/com/blist/metrics/impl/queue/MetricsBufferSpec.scala
```diff
@@ -0,0 +1,51 @@
...
+      val c1 = metricsBuffer.popAll();
+      c1.add(new MetricsBag("some_entity_id", metrics, System.currentTimeMillis()))
+      assert(metricsBuffer.size() == 0, "Metrics Buffer is exposing internal representation.")
+    }
+
+    // TODO Write much more test
```
jacobsi: Not worth recording as a TODO. This is nearly always true.

balboa-client-jms/src/main/java/com/blist/metrics/impl/queue/MetricJmsQueue.java
mhotan: I will remove the random TODO.  After looking at it closely, I do not agree with the naming and how this particular class named its static member.

As far as the MetricJmsQueueNotSingleton class, I am not a big fan of it either.  It was an implementation I walked in on and was not wanting to change the implementation/interface much.

balboa-agent/src/main/java/com/socrata/balboa/agent/MetricConsumer.java
jacobsi: You could throw an exception for the error case and return null for EOF. That would be more in line with how I would expect it to work as well.

balboa-client-jms/src/main/java/com/blist/metrics/impl/queue/MetricJmsQueue.java
mhotan: MetricJmsQueue is the base class.  It does not inherit directly from any other implementations.  It does have probably one to many layers of interfaces.

balboa-client-jms/src/main/java/com/blist/metrics/impl/queue/MetricJmsQueueNotSingleton.java
mhotan: See above

balboa-agent/src/main/java/com/socrata/balboa/agent/MetricConsumer.java
jacobsi: I know you didn't name that function, but `grovel` is a terrible name. `readRecord` seems to succinctly summarize what it is doing?

What is that big while loop with a label even doing in there? I can't see it's purpose.

It seems like:
0xfe = End of Record
0xff = Beginning of Record

??

balboa-agent/src/main/java/com/socrata/balboa/agent/MetricConsumer.java
```diff
@@ -71,6 +71,24 @@ public void run() {
         }
     }
 
+    /**
+     * Closing the Metric Consumer is effectively a delegation method that allows the Metric Consumers
+     * internal resources to close and clean up (If necessary).
```
jacobsi: This is pretty much what all .close() methods do. You don't even have to write this if you are feeling lazy.

balboa-agent/src/main/java/com/socrata/balboa/agent/MetricConsumer.java
```diff
@@ -71,6 +71,24 @@ public void run() {
...
+     *
+     * @throws IOException
```
jacobsi: If you aren't going to explain anything extra about this, you might as well delete this line. I can see this info in the type signature.

balboa-agent/src/main/java/com/socrata/balboa/agent/MetricConsumer.java
```diff
@@ -18,7 +18,7 @@
  * NOTE: Copied and Pasted from another existing project.
  * TODO: Deprecate this Concurrent Modification Exception prone class.
```
jacobsi: Seems like this note no longer applies.

balboa-agent/src/main/java/com/socrata/balboa/agent/MetricConsumer.java
```diff
@@ -18,7 +18,7 @@
  * NOTE: Copied and Pasted from another existing project.
```
jacobsi: Not important, unless it reference where it was taken from.

balboa-agent/src/main/java/com/socrata/balboa/agent/MetricConsumer.java
```diff
@@ -18,7 +18,7 @@
...
  * TODO: Deprecate this Concurrent Modification Exception prone class.
  */
```
jacobsi: Since this is a pretty central class, and the name doesn't provide much insight into it's true nature (it's both a consumer and a producer, consuming from local files, sending to a remote AMQ queue), a quick summary here saying how it runs as a separate thread and reads metrics files from the files in the directory tree, buffering them before sending them on to AMQ.

balboa-agent/src/main/java/com/socrata/balboa/agent/MetricConsumer.java
jacobsi: If you throw an exception for all the error conditions in grovel, and return null for all the EOF conditions, your while loop becomes:

Map<String, String> record;
while ((record = grovel(stream)) != null) {
...
}

balboa-agent/src/main/scala/com/socrata/balboa/agent/BalboaAgent.scala
```diff
@@ -52,15 +40,29 @@ object BalboaAgent extends App with Config with LazyLogging {
...
+  logger.debug(s"Created Activemq Connection $amqConnection")
+
+  // TODO (Pre req: Java 8) Use Java Duration class.
+  val future: ScheduledFuture[_] = scheduler.scheduleWithFixedDelay(new Runnable {
+    override def run(): Unit = {
+      // Wrapping the Metric Consumer inside an anonymous inner function makes the metric consumer more ephemeral
+      // This will require a little more work from the JVM to create and cleanup but that is what it is designed to do.
+      val mc = new MetricConsumer(dataDir, new MetricJmsQueueNotSingleton(amqConnection, activemqQueue))
+      logger.debug(s"Attempting to run Metric Consumer $mc")
+      mc.run()
```
jacobsi: try {
    ...
} finally {
    mc.close()
}

might be appropriate here.

balboa-agent/src/main/scala/com/socrata/balboa/agent/BalboaAgent.scala
```diff
@@ -52,15 +40,29 @@ object BalboaAgent extends App with Config with LazyLogging {
...
+    case (Some(user), Some(password)) => ActiveMQConnection.makeConnection(user, password, activemqServer)
+    case _ => ActiveMQConnection.makeConnection(activemqServer)
   }
-  val future: ScheduledFuture[_] = scheduler.scheduleWithFixedDelay(mc, INITIAL_DELAY, period, TimeUnit.MILLISECONDS)
+  logger.debug(s"Created Activemq Connection $amqConnection")
+
+  // TODO (Pre req: Java 8) Use Java Duration class.
+  val future: ScheduledFuture[_] = scheduler.scheduleWithFixedDelay(new Runnable {
+    override def run(): Unit = {
+      // Wrapping the Metric Consumer inside an anonymous inner function makes the metric consumer more ephemeral
```
jacobsi: I don't understand what this means. What problem are you fixing?

balboa-client/src/main/java/com/blist/metrics/impl/queue/MetricsBuffer.java
mhotan: First answer, yes.  

Second answer, yes.  Metric event logs can be read from disk, transmitted, and process crashes with a non IOException and the consumer fails to mark the file as broken and the process restarts later.

balboa-client/src/main/java/com/blist/metrics/impl/queue/MetricsBuffer.java
mhotan: Is that much of an issue if the Collection is not a reference to the underlying member to MetricsBuffer.  Is there a lot of harm to passing them an interface to Collection when manipulations to that collection will not effect the buffer at all?

mhotan: That being said there are mutability vulnerabilities all over the place.  Especially relevant with Metric and Metrics class.  But those would not be alleviated with that returning a List or a Collection.

balboa-agent/src/main/java/com/socrata/balboa/agent/MetricConsumer.java
mhotan: I am struggling to understand what the big while true loop with the label `outer` is doing.  I am tempted to remove it to identify what it is doing.

balboa-agent/src/main/scala/com/socrata/balboa/agent/BalboaAgent.scala
mhotan: I found that having an internal buffer that flushes based off of size could result in metric loss if the buffer fills to anything buts its flushing capacity and then gets GCed.  In order to handle that case and preserve a simple abstraction using a well known abstraction I had Metric Consumer implement AutoCloseable.  When MetricConsumer.close is invoked the buffer attempts to flush.  So here I am wrapping MetricConsumer run and close in one Runnable anonymous class to ensure that all the metric messages are sent to AMQ.

balboa-client/src/test/scala/com/blist/metrics/impl/queue/MetricsBufferSpec.scala
mhotan: Yep that was the wrong terminology.  I am going to change it to reflect that it is attempting to test whether or not MetricBuffer inadvertently reveal underlying representation.

balboa-client-kafka/src/main/java/com/socrata/balboa/impl/MetricLoggerJava.java
```diff
@@ -63,4 +63,9 @@ private MetricLoggerJava(String topic) {
     public void create(IdParts entity, IdParts name, long value, long timestamp, Metric.RecordType recordType) {
         this.logger.logMetric(entity.toString(), name.toString(), value, timestamp, recordType);
     }
+
+    @Override
+    public void close() throws Exception {
+        // TODO Don't know how or what it means to shut this down.
```
jacobsi: If you have to override the close method (it's abstract or from an interface), the comment should read something like "There is no cleanup activity to be completed."

As long as you have closed any member variables of this class instance, and freed any resources allocated to this class instance, close has accomplished it's objective.

balboa-client-jms/src/main/java/com/blist/metrics/impl/queue/MetricJmsQueue.java
jacobsi: So you mean there is no super.close to chain to? It's abstract or from an interface?

mhotan: AbstractJavaMetricQueue is an abstract class which indirectly implements MetricQueue.  Therefore, to my knowledge, no method to chain to.

jacobsi:
lgtm

PR #67, opened 2016-02-03
EN-2457: Publish balboa-client with Scala 2.11

mhotan:
* Updated syntax to Scala 2.11 standards
* Update build scripts.

### DIFF
--- a/balboa-service-kafka/src/main/scala/com/socrata/balboa/service/kafka/KafkaConsumerCLIBase.scala
+++ b/balboa-service-kafka/src/main/scala/com/socrata/balboa/service/kafka/KafkaConsumerCLIBase.scala
@@ -27,7 +27,6 @@ trait KafkaConsumerCLIBase[K,M] extends App {
   //////////////////////////////////////////////////////////////////////////////////////////
   /////// Running the Application
 
-  super.main(args)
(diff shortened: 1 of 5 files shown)


balboa-service-kafka/src/main/scala/com/socrata/balboa/service/kafka/KafkaConsumerCLIBase.scala
```diff
@@ -28,33 +27,31 @@ trait KafkaConsumerCLIBase[K,M] extends App {
...
-      g.stop()
-    } catch {
-      case e: Exception => Log.warn("Unable to stop consumer group. ", e)
-      case t: Throwable => Log.warn(s"Unknown error while stopping consumer group.", t)
-    })
-
-    Log.info(s"Starting $g...")
-    g.start()
-  }
+  super.main(args)
```
jacobsi: Do you still need this? If you are not overriding main, then the super.main will be invoked naturally.

mhotan: Good catch.

raganw:
With Jacob's comment addressed, :shipit:

PR #68, opened 2016-02-06
EN-2400 Immutable Files

mhotan:
Added support for immutable files in metric consumer and metric file
queue.  This allows containers to write to files and declare them to
be immutable once they are completed.


mhotan:
@smooshy This is the commit that will begin accepting files ending with `.completed` or `.COMPLETED`.

balboa-agent/src/main/scala/com/socrata/balboa/agent/metrics/BalboaAgentMetrics.scala
```diff
@@ -83,17 +83,23 @@ object BalboaAgentMetrics {
...
+    * @return Success(Metric) if successfully create the metric, False otherwise
     */
-  def directorySize(metricName: String, directory: File): Try[Metric] = directory match {
+  def numFiles(metricName: String, directory: File, filter: Option[FileFilter]): Try[Metric] = directory match {
     case f: File if directory.exists() && directory.isDirectory =>
-      Success(registry.register(MetricRegistry.name(name, metricName, "size"),
+      Success(registry.register(MetricRegistry.name(name, metricName, "num", "files"),
         new CachedGauge[Int](5, TimeUnit.MINUTES) { //
-        override def loadValue(): Int = FileUtils.getDirectories(directory).size()
+        override def loadValue(): Int = FileUtils.getDirectories(directory).map(dir =>
```
mhotan: I am a little worried about FIle IO race conditions that could cause unexpected exceptions.  Right now we are assuming the directory does not get deleted prior to calling list files.

balboa-client/src/main/java/com/blist/metrics/impl/queue/MetricFileQueue.java
```diff
@@ -72,18 +84,26 @@ private void open() throws IOException {
...
-        fileStream = new FileOutputStream(new File(directory, logName), true);
-        stream = new BufferedOutputStream(fileStream);
-        reopenTime = now + MetricQueue$.MODULE$.AGGREGATE_GRANULARITY();
+        file = new File(directory, logName);
+        stream = new BufferedOutputStream(new FileOutputStream(new File(directory, logName), true));
+        reopenTime = now + this.reopenInterval;
         metricCount = 0;
     }
 
+    private void renameFile(File file) {
```
housepage: Let's make this more descriptive like: renameFileWithImmutableExtension or markFileAsImmutable? 

Nit: Also, can we return the result from the rename from the function and do the logging outside the function?

balboa-client/src/main/java/com/blist/metrics/impl/queue/MetricFileQueue.java
```diff
@@ -72,18 +84,26 @@ private void open() throws IOException {
         this.directory.mkdirs();
 
         String logName = basename + ".data";
-        fileStream = new FileOutputStream(new File(directory, logName), true);
-        stream = new BufferedOutputStream(fileStream);
-        reopenTime = now + MetricQueue$.MODULE$.AGGREGATE_GRANULARITY();
+        file = new File(directory, logName);
```
housepage: Why assign the file variable here and then not use it in the next line?

balboa-agent/src/main/scala/com/socrata/balboa/agent/metrics/BalboaAgentMetrics.scala
housepage: We should add exception handling here for that contingency and then add another metric for that.

balboa-client/src/main/java/com/blist/metrics/impl/queue/MetricFileQueue.java
mhotan: Whoops that is a mistake.  I should be reusing file in the following line.  We also need a reference to it on close

balboa-agent/src/main/scala/com/socrata/balboa/agent/metrics/BalboaAgentMetrics.scala
mhotan: How about a log message we can scoop up later?  I think we have enough metricception going on.

housepage: Fine by me :+1:

housepage:
:two_men_holding_hands: Beautiful!

mhotan:
Hmm for some reason the test failed in jenkins.  I am rerunning them to see if I can recreate it.  If I cannot recreate it then Ill go ahead and merge it.  There are a few test that are dependent on time.  The combination of those test and the resources the on the test executor can result in one or two test timing out.

mhotan:
Unable to recreate on our jenkity jenkins.  Travis passed so I am going to go ahead and merge it.

PR #69, opened 2016-02-23
Fixed Incorrect Buffer Capacity Def.

mhotan:
* The buffercapacity in MetricJmsQueueNotSingleton was not being
 utilized correctly.  This change fixes that.

NOTE: MetricJmsQueueNotSingleton should be refactored into
MetricJmsQueue.  This changes the interface and I deemed that it should
be released with this patch.


housepage:
:two_men_holding_hands: :balloon: :rocket:

PR #70, opened 2016-02-26
Actually parse args properly for set of uri for amq

jreichhold:
Improve parsing of AMQ uri and add timeouts if not specified. Also allows comma-seperated list of URI vs space-seperated set

### DIFF
--- a/balboa-service-jms/src/main/java/com/socrata/balboa/jms/BalboaJms.java
+++ b/balboa-service-jms/src/main/java/com/socrata/balboa/jms/BalboaJms.java
@@ -8,7 +8,8 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
-import java.util.Arrays;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Properties;
 
 public class BalboaJms {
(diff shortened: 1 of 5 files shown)


balboa-service-jms/src/main/java/com/socrata/balboa/jms/BalboaJms.java
```diff
@@ -24,9 +25,21 @@ static Integer parseThreads(String[] args)
...
-    static String[] parseServers(String[] args)
+    static List<String> parseServers(String[] args)
     {
-        return Arrays.copyOfRange(args, 1, args.length - 1);
+        List<String> servers = new ArrayList<>();
+        for(int i=1; i< args.length-1; i++){
+            String[] srvs = args[i].split(",+(?![^\\(]*\\))");
+            for(String s:srvs) {
+                if(s.indexOf("?")<0) {
+                    s += "?soTimeout=15000&soWriteTimeout=15000";
```
mhotan: I will make those values configurable at a later date and time

mhotan:
Looks good to me :shipit:

balboa-service-jms/src/main/java/com/socrata/balboa/jms/BalboaJms.java
```diff
@@ -49,12 +65,14 @@ public static void main(String[] args) throws Exception
             System.exit(1);
         }
 
+        Properties p = new Properties();
+        p.load(BalboaJms.class.getClassLoader().getResourceAsStream("config/config.properties"));
```
jacobsi: getResourceAsStream can return a null. If it does, it would be better if the message said "'config/config.properties' not found on the classpath" instead of NullPointerException.

PR #71, opened 2016-02-26
Fix assembly with spring from latest activemq

jreichhold:


mhotan:
Build config change so :shipit:

PR #72, opened 2016-03-29
Balboa HTTP build fix

mhotan:
Cross scala versions are now set at sub projects that require being
published to repo.socrata.com.


TranquilMarmot:
:sunglasses: :shipit:

PR #77, opened 2016-04-12
Fix inconsistent error reporting.

jacobsi:
When loading from the config files, error reporting wasn't going to the
logs, some error messages were wrong, some were misleading.

### DIFF
--- a/balboa-common/src/main/java/com/socrata/balboa/metrics/config/Configuration.java
+++ b/balboa-common/src/main/java/com/socrata/balboa/metrics/config/Configuration.java
@@ -11,11 +11,6 @@ import java.util.List;
 import java.util.Properties;
 
 public abstract class Configuration extends Properties {
-
-    /**
-     * TODO: Can't figure out why we are mutex locking Configuration Queries.
-     */
-
     public static Configuration instance;
     private List<Period> supportedPeriods;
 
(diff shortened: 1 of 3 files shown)


jacobsi:
please retest this

jacobsi:
retest this please

balboa-common/src/main/java/com/socrata/balboa/metrics/config/Configuration.java
```diff
@@ -45,15 +40,15 @@ public static synchronized Configuration get() throws IOException {
...
+    public synchronized void setSupportedTypes(List<Period> supportedPeriods) {
         this.supportedPeriods = supportedPeriods;
     }
 
     public synchronized List<Period> getSupportedPeriods() {
         if (supportedPeriods == null) {
             String[] types = getProperty("balboa.summaries").split(",");
 
-            supportedPeriods = new ArrayList<Period>(types.length);
+            supportedPeriods = new ArrayList<>(types.length);
```
housepage: Why get rid of the generic type here?

jacobsi: Java 7 syntax diamond syntax. A generic type for ArrayList not required if it can be inferred from the type of the variable. In this case, Java knows `supportedPeriods` has the generic type `Period`, so I don't have to specify the generic type of the `ArrayList`.

Since it's not necessary to specify `Period` with `ArrayList`, Intellij highlights it as a warning, and I like to clear as many warnings as possible. A few times I've run across real errors as I do it.

housepage:
:shipit:

PR #85, opened 2016-04-23
EN-4888 Add transport listener to log more errors.

jacobsi:


raganw:
Looks good to me :shipit:

PR #87, opened 2016-04-25
Increase ActiveMQ client library logging.

jacobsi:
When there are connection related problems, the ActiveMQ client
libraries handle the connection/reconnection, timeout and failover logic
invisibly. This change adds logging configuration to increase the level
of ActiveMQ logging specifically around connections and failovers. When
experimenting on my local machine, the amount of logging introduced by
this change is something on the order of 1-2 lines every few seconds,
when connection problems are happening.


TranquilMarmot:
👍  lgtm

PR #90, opened 2016-04-26
Enable cross compiling.

jacobsi:


TranquilMarmot:
That easy, huh? :shipit:

jacobsi:
Kind of. Seems there is some weird configuration for this project and I have to cross compile all the sub-projects to make cross compiling work. I think I should be able to cross compile the specific sub-projects, but I'm going to do this to get everyone unblocked and look deeper into sbt later.

PR #91, opened 2016-04-26
Use a LoggingScheduledExecutor

rozap:
* The regular one halts the subsequent runs if an exception is thrown
* This one at least logs that something went wrong

Refs EN-5238

### DIFF
--- a/balboa-agent/src/main/scala/com/socrata/balboa/agent/BalboaAgent.scala
+++ b/balboa-agent/src/main/scala/com/socrata/balboa/agent/BalboaAgent.scala
@@ -109,6 +109,8 @@ object BalboaAgent extends App with Config with StrictLogging {
       try {
         logger debug s"Attempting to run Metric Consumer $mc"
         mc.run() // Recursively reads all metric data files and writes them to a queue.
+      } catch {
+        case e: Exception => logger.error("Metric consumer failed", e)
       } finally {
         logger debug s"Attempting to close Metric Consumer $mc"
         mc.close() // Release all associated resources.


balboa-agent/src/main/scala/com/socrata/balboa/agent/LoggingScheduledExecutor.scala
```diff
@@ -0,0 +1,26 @@
...
+import org.slf4j.LoggerFactory
+
+class LoggingScheduledExecutor(poolSize: Int) extends ScheduledThreadPoolExecutor(poolSize) {
+  override def scheduleAtFixedRate(runnable: Runnable, delay: Long, period: Long, unit: TimeUnit): ScheduledFuture[_] = {
+    val wrapped = new LogRunnableException(runnable)
+    super.scheduleAtFixedRate(wrapped, delay, period, unit)
+  }
+
+  private class LogRunnableException(underlying: Runnable) extends Runnable {
+    val logger: Logger = Logger(LoggerFactory getLogger getClass.getName)
```
jacobsi: Why not use `with StrictLogging`?

balboa-agent/src/main/scala/com/socrata/balboa/agent/LoggingScheduledExecutor.scala
```diff
@@ -0,0 +1,26 @@
...
+    override def run(): Unit = {
+      try {
+        underlying.run()
+      } catch {
+        case e: Exception =>
+          logger.error("Underlying runnable threw an exception", e)
+      }
+    }
+  }
+}
```
jacobsi: Why are you building a new class? Why isn't it sufficient to add a catch clause to the runnable in BalboaAgent?

rozap:
Code coverage failed; I'm not sure that a sane way exists on the JVM to test log output? Redirecting stdout/stderr would be problematic for a multithreaded test execution. Also I really have no idea what the logging situation is here; there are a number of different logging backends it looks like. 

It seems like we'll have to do nastier things to get the coverage up, rather than just adding nocover to this block. 

Any thoughts?

balboa-agent/src/main/scala/com/socrata/balboa/agent/LoggingScheduledExecutor.scala
rozap: ok

jacobsi:
lgmt

PR #94, opened 2016-05-07
Bump down "Using entity/row key" log message to debug level from info.

znep:
balboa-jms is currently logging ~10 gigs of logs a day, and ~80%
of them are from this one log line, which doesn't seem overly useful
or necessary to log at non-debug levels.


jacobsi:
lgmt

PR #95, opened 2016-05-10
Address cross compiling problems

jacobsi:
balboa-client-* must be cross compiled for Scala 2.10 and 2.11 because
there are some projects of each version consuming them.

balboa-http can not be cross compiled for Scala 2.11 because the
socrata-http dependency is only in Scala 2.10.

The way our build system works it invokes all sbt tasks from the root
project:

    sbt +balboa-client-dispatcher/compile

When `sbt` is invoked this way, it uses the crossScalaVersions setting
from the root project, not from the project being compiled.

In the end we fixed this by munging the build job on Jenkins to pass an
additional parameter to sbt, thusly:

    sbt 'set crossScalaVersions = List("2.10.6", "2.11.7")' \
        +balboa-client-dispatcher/compile

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,7 +1,39 @@
-// TODO Delete... I don't think this is actually be used.
-scalaVersion := "2.10.6"
+//
+// Cross compiling situation
+//
+// As things stand, the balboa-clients need to be cross compiled because they
+// are published as libraries and consumed by both Scala 2.10 and Scala 2.11
+// projects.
+//
+// balboa-http can not be cross compiled because of it's dependency on the
+// socrata-http library. socrata-http is compiled for Scala 2.10 in the version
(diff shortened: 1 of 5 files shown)


build.sbt
```diff
@@ -1,7 +1,39 @@
...
-scalaVersion := "2.10.6"
+//
+// Cross compiling situation
+//
+// As things stand, the balboa-clients need to be cross compiled because they
+// are published as libraries and consumed by both Scala 2.10 and Scala 2.11
+// projects.
+//
+// balboa-http can not be cross compiled because of it's dependency on the
+// socrata-http library. socrata-http is compiled for Scala 2.10 in the version
```
TranquilMarmot: Ouch!

TranquilMarmot:
lgtm!

PR #99, opened 2016-05-18
Fix style, add method return types.

jacobsi:


balboa-client/src/main/scala/com/socrata/balboa/impl/HashMapBuffer.scala
```diff
@@ -8,8 +8,8 @@ import com.socrata.metrics.MetricQueue
 import com.socrata.metrics.components.{BufferComponent, BufferItem, MessageQueueComponent}
 import org.slf4j.LoggerFactory
 
+import collection.mutable
```
raganw: Is there a reason to not include the `scala.` at the beginning of the package name here?

balboa-common/src/main/scala/com/socrata/balboa/metrics/MetricIdentifiers.scala
```diff
@@ -7,28 +7,28 @@ import java.util.UUID
...
+    MetricIdParts( getParts.map { p:MetricIdPart => if (p.isUnresolved) part else p }:_* )
   }
-  def hasPart(part:MetricIdPart) = {
+  def hasPart(part:MetricIdPart): Boolean = {
     getParts != null && getParts.exists {
       e:MetricIdPart => e == part
     }
   }
-  def isUnresolved() = {
+  def isUnresolved: Boolean = {
```
raganw: In other files, you have retained the empty parentheses.  We probably want to choose one way or the other.

balboa-core/src/main/scala/com/socrata/balboa/metrics/data/impl/Cassandra11DataStore.scala
```diff
@@ -50,7 +50,7 @@ class Cassandra11DataStore(queryImpl:Cassandra11Query = new Cassandra11QueryImpl
     entities("")
   }
 
-  def getValidGranularity(period:Period) = {
+  def getValidGranularity(period:Period): Period = {
```
raganw: Missing space after `:` in the parameters list.

raganw:
A few comments/questions, but otherwise looks good.

balboa-client/src/main/scala/com/socrata/balboa/impl/HashMapBuffer.scala
jacobsi: Don't have to for Scala packages. I didn't do it on purpose, Intellij auto imported. I kind of like saying 'scala', so I'll fix that.

balboa-common/src/main/scala/com/socrata/balboa/metrics/MetricIdentifiers.scala
jacobsi: By convention, the () parenthesis in Scala is suppose to mean a method with a side effect. So, things like `close()` come with (), and things like `getX` are without.

I didn't religiously fix up all the files I touched, but if I was in one, I tried to clear additional Intellij warnings. I assume that's why I removed these ones, as Intellij should detect this as a getter.

PR #100, opened 2016-05-18
Disable kafka related tests.

jacobsi:
These tests are unstable and inconsistent on Jenkins, which is a real
barrier to pull requests. More to the point, we no longer support Kafka.
There is future work to delete this code:

    https://socrata.atlassian.net/browse/EN-6069

For the moment, just disabling it.


housepage:
:shipit:

PR #101, opened 2016-05-19
Fix error when requesting /

jacobsi:
An error was occurring when requesting / on balboa-http. This change
fixes that error and adds basic integration test coverage to balboa-http
to verify the fix did not cause any regressions.


TranquilMarmot:
lgtm :shipit:

PR #105, opened 2016-05-31
Fix balboa-agent style issues.

jacobsi:
Correct all the style violations in the balboa-agent package and enable
build checks for that package.


raganw:
That "magic number" rule is a killer.  At any rate, looks good. :shipit:

PR #108, opened 2016-06-08
Jacobsi/rename balboa schema

jacobsi:


raganw:
Looks good to me - Squash and :shipit:

PR #110, opened 2016-06-14
EN-7053 - Add support for DC in C* configuration

raganw:
If Cassandra datacenter is specified in configuration properties, apply
it as a DC local configuration.  Otherwise, behave as it has previously.

### DIFF
--- a/balboa-core/src/main/scala/com/socrata/balboa/metrics/data/impl/Cassandra11Util.scala
+++ b/balboa-core/src/main/scala/com/socrata/balboa/metrics/data/impl/Cassandra11Util.scala
@@ -95,6 +95,7 @@ object Cassandra11Util extends StrictLogging {
 
     val seeds = conf.getProperty("cassandra.servers")
     val keyspace = conf.getProperty("cassandra.keyspace")
+    val datacenter = Option(conf.getProperty("cassandra.datacenter"))
     val sotimeout = conf.getProperty("cassandra.sotimeout").toInt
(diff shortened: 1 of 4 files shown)


balboa-core/src/main/scala/com/socrata/balboa/metrics/data/impl/Cassandra11Util.scala
```diff
@@ -103,16 +104,25 @@ object Cassandra11Util extends StrictLogging {
...
+      .setTimeoutWindow(sotimeout)
+      .setMaxConnsPerHost(connections)
+      .setSeeds(seeds)
+
+    // Set local DC as side-effect if specified in configuration.
+    // If the local datacenter is specified it will limit the driver
+    // to only make connections to the Cassandra nodes in the datacenter
+    // and prevent this service from unintentionally reaching across a
+    // VPN to connect to a Cassandra node.
+    datacenter.foreach(connectionPoolConfiguration.setLocalDatacenter)
```
jacobsi: Do you have some reason to believe that it is possible to set multiple local datacenters? When I look at the implementation of `setLocalDatacenter`, it doesn't seem to accumulate them, it seems to just set one object level field. And `getLocalDatacenter` just returns a String.

jacobsi: Really like the way it only sets the local datacenter if the config setting is found.

jacobsi:
lgtm

PR #112, opened 2016-06-15
Add badcall integration tests and protobuf response code tests to balboa-http

cramertj:
Note: protobuf response code tests purposefully left out of `/metrics/*/series` queries because the series endpoint does not support protobuf.


jacobsi:
lgtm

PR #117, opened 2016-06-16
EN-7066 - Convert Balboa HTTP from Socrata HTTP to Scalatra

cramertj:


balboa-http/src/main/scala/com/socrata/balboa/server/Main.scala
```diff
@@ -2,47 +2,79 @@ package com.socrata.balboa.server
...
+import org.json4s.{DefaultFormats, Formats}
+import org.scalatra.servlet.ScalatraListener
+import org.scalatra.json.JacksonJsonSupport
 
-class Main
+import scala.collection.JavaConverters._
 
-object Main extends App with StrictLogging {
-  val DefaultPort = 9012
+class MainServlet extends ScalatraServlet with StrictLogging with JacksonJsonSupport {
```
jacobsi: This can go in it's own file. Not required for Scala, but this is big enough to make it worth while. It's generally easier to find things where the class name corresponds to the file name.

balboa-http/src/main/scala/com/socrata/balboa/server/MainServlet.scala
```diff
@@ -0,0 +1,67 @@
...
+
+  val versionString = BuildInfo.toJson
+
+  def getAccepts(req: HttpServletRequest): Seq[String] = {
+    req.getHeaders("accept").asScala.toSeq
+  }
+
+  get("/*") {
+    contentType = "application/json"
+    NotFound(Error(404, "Not found."))
```
jacobsi: NotFound is a 404. Is it necessary to specify anything? What about:

    NotFound()

or:

    NotFound(reason = "Not found.")

cramertj: The error supplies the JSON body of the response, which in this case is:
```json'
{
  "error": 404,
  "message": "Not found."
}
```

balboa-http/src/main/scala/com/socrata/balboa/server/MainServlet.scala
```diff
@@ -0,0 +1,73 @@
...
+
+  get("/metrics/:entityId") {
+    val entityId = params("entityId")
+    val response = MetricsRest.get(entityId, params, getAccepts(request))
+    contentType = response.contentType
+    response.result
+  }
+
+  // Catch the extra paths to match behavior of old API
+  get("/metrics/:entityId/*") {
```
jacobsi: Since this method is exactly the same as `get("/metrics/:entityId")`, can you combine them? How about something like this, http://scalatra.org/2.4/guides/http/routes.html#toc_295, or a regex pattern matcher if necessary?

balboa-http/src/main/scala/com/socrata/balboa/server/Main.scala
```diff
@@ -1,48 +1,14 @@
...
-import com.socrata.balboa.metrics.config.Configuration
-import com.socrata.balboa.server.rest.{EntitiesRest, MetricsRest, VersionRest}
-import com.socrata.http.routing.{HttpMethods, SimpleRoute, SimpleRouter}
-import com.socrata.http.server.implicits._
-import com.socrata.http.server.responses._
-import com.socrata.http.server.{Filter, HttpResponse, Service, SimpleFilter, SocrataServerJetty}
-import com.typesafe.scalalogging.slf4j.StrictLogging
-import org.apache.log4j.PropertyConfigurator
+object Main extends App {
+  val logger = LoggerFactory.getLogger(getClass)
```
jacobsi: You should be able to use:

    with StrictLogging

instead of an explicit logger in the class.

balboa-http/src/main/scala/com/socrata/balboa/server/ResponseWithType.scala
```diff
@@ -0,0 +1,23 @@
...
+  def unacceptable: ResponseWithType =
+    ResponseWithType(json, NotAcceptable(Error(406, "Not acceptable.")))
+
+  def required(parameter: String): ResponseWithType =
+    ResponseWithType(json, BadRequest(Error(400, s"Parameter $parameter required.")))
+
+  def malformedDate(parameter: String): ResponseWithType =
+    ResponseWithType(json, BadRequest(Error(400, "Unable to parse date " + parameter)))
+
+  def br(parameter: String, msg: String): ResponseWithType =
```
jacobsi: Full name for this method. `badRequest`

balboa-http/src/main/scala/com/socrata/balboa/server/rest/MetricsRest.scala
```diff
@@ -143,8 +111,7 @@ object MetricsRest {
   }
 
   // ok, so this isn't actually correct.  Screw correct, I just want this to work.
```
jacobsi: Do you have any idea what this comment might refer to?

project/BalboaHttp.scala
```diff
@@ -2,20 +2,27 @@ import Dependencies._
 import sbt.Keys._
 import sbt._
 import scoverage.ScoverageSbtPlugin
+//import com.earldouglas.xwp.JettyPlugin
```
jacobsi: If you don't need this, you can delete it.

project/BalboaHttp.scala
```diff
@@ -2,20 +2,27 @@ import Dependencies._
...
+import org.scalatra.sbt.ScalatraPlugin
 
 object BalboaHttp {
-  lazy val settings: Seq[Setting[_]] = BuildSettings.projectSettings ++ Seq(
+  lazy val settings: Seq[Setting[_]] =
+    //(project in file(".")).enablePlugins(JettyPlugin).settings ++
```
jacobsi: Same here. If you don't need this, you can delete it.

balboa-http/src/main/scala/com/socrata/balboa/server/rest/MetricsRest.scala
cramertj: Nope. Seeing as we're planning to deprecate the protobuf support, this code will probably disappear soon.

project/BalboaHttp.scala
cramertj: That was a mistake. I've reverted it.

balboa-http/src/main/scala/com/socrata/balboa/server/rest/MetricsRest.scala
jacobsi: Delete the comment. It doesn't have any content so it doesn't have any value.

project/Balboa.scala
```diff
@@ -20,6 +21,7 @@ object Balboa extends Build {
   lazy val balboaCore = project("balboa-core", BalboaCore, balboaCommon)
 
   lazy val balboaHttp = project("balboa-http", BalboaHttp, balboaCore)
+    .enablePlugins(JettyPlugin)
```
cramertj: @jacobsi I've added it here instead.

project/BalboaHttp.scala
cramertj: To run from the command line, I use `sbt balboa-http/assembly` from `balboa`.

balboa-http/src/main/scala/com/socrata/balboa/server/ResponseWithType.scala
```diff
@@ -9,15 +9,21 @@ object ResponseWithType {
   val json = "application/json; charset=utf-8"
   val protobuf = "application/x-protobuf"
 
+  val BadRequestCode = 400
+  val InternalErrorCode = 500
+  val NotFoundCode = 404
+  val NotAcceptableCode = 406
```
jacobsi: Why not use:

    org.apache.http.HttpStatus.SC_BAD_REQUEST
    org.apache.http.HttpStatus.SC_INTERNAL_SERVER_ERROR
    org.apache.http.HttpStatus.SC_NOT_FOUND
    org.apache.http.HttpStatus.SC_NOT_ACCEPTABLE

Or org.eclipse.jetty.http.HttpStatus has them too.

balboa-http/src/main/scala/com/socrata/balboa/server/rest/EntitiesRest.scala
```diff
@@ -18,14 +18,18 @@ object EntitiesRest {
...
     val it = ds.entities().asScala.filter(predicate)
     // backwards-compatibility: -1 == no limit
     val limitString = params.getOrElse("limit", "-1")
     val limit = Try(limitString.toInt).recover({
       case err: NumberFormatException =>
-        return ResponseWithType(json, BadRequest(Error(400, "Unable to parse limit : " + JString(err.toString).toString.drop(1).dropRight(1))))
+        return ResponseWithType(json,
+          BadRequest(Error(
+            BadRequestCode,
+            "Unable to parse limit : " + JString(err.toString).toString.drop(1).dropRight(1))))
```
jacobsi: If I've understood this line:

    err.toString

produces:

    java.lang.NumberFormatException: For input string: "abc"

Then:

    JString(err.toString).toString

produces:

    "java.lang.NumberFormatException: For input string: \"abc\""

Then:

    JString(err.toString).toString.drop(1)

produces:

    java.lang.NumberFormatException: For input string: \"abc\""

Then:

    JString(err.toString).toString.drop(1).dropRight(1)

produces:

    java.lang.NumberFormatException: For input string: \"abc\"

which is almost what was there at the beginning. Which seems like an awful lot of work for what will really be a terrible error message in the end.

How about something like:

    s"The value supplied for 'limit' is '$limitString', which can not be parsed as an integer."

project/BalboaHttp.scala
```diff
@@ -2,19 +2,15 @@ import Dependencies._
...
 
 object BalboaHttp {
   lazy val settings: Seq[Setting[_]] =
-    //(project in file(".")).enablePlugins(JettyPlugin).settings ++
     BuildSettings.projectSettings ++ Seq(
     libraryDependencies <++= scalaVersion {libraries(_)},
     ScoverageSbtPlugin.ScoverageKeys.coverageMinimum := 0
   ) ++ ScalatraPlugin.scalatraFullSettings
 
-  // balboa-http can not be cross compiled because of the dependency on
```
jacobsi: There is another comment in the build.sbt that can be fixed up too. You should only need to remove the part about balboa-http being incompatible with Scala 2.11. Leave the rest. It can be cleaned up, as necessary, when we update the repo to use Scala 2.11.

balboa-http/src/main/scala/com/socrata/balboa/server/rest/EntitiesRest.scala
jacobsi: If you do get rid of JString, could you check if rojoma json is a necessary dependency of balboa-http?

cramertj: The JString logic was there before I touched the codebase, and used to exist in several other places as well. I believe the sole purpose of it was to escape the String into a valid JSON String (escaping quotes, special characters, and the like). I wouldn't be surprised if that's unnecessary now I'm using json4s to automatically convert my case class to and from JSON. I'll remove it pending investigation.

balboa-http/src/main/scala/com/socrata/balboa/server/rest/EntitiesRest.scala
```diff
@@ -1,39 +1,43 @@
...
-      case Left(err) =>
-        return BadRequest ~> ContentType("application/json") ~> Content("""{"error": 400, "message": "Unable to parse limit : """ + JString(err).toString.drop(1).dropRight(1) + """"}""")
-    }
+    val limitString = params.getOrElse("limit", "-1")
+    val limit = Try(limitString.toInt).recover({
+      case err: NumberFormatException =>
+        return ResponseWithType(json,
+          BadRequest(Error(
+            BAD_REQUEST_400,
+            "Unable to parse limit : " + err.toString)))
```
jacobsi: Not err.toString. That's implementation dependent detail. And in theory the output of toString can change at any time. (It never does though.) Just a message that indicates the value can't be parsed as an int is sufficient.

cramertj: You're fine with making that a breaking change?

jacobsi: Yes, I'm okay with changing an error message. People who parse error messages are on their own.

jacobsi:
lgtm

PR #118, opened 2016-06-16
EN-7053 - Add logging of datacenter if specified

raganw:


balboa-core/src/main/scala/com/socrata/balboa/metrics/data/impl/Cassandra11Util.scala
```diff
@@ -100,6 +100,7 @@ object Cassandra11Util extends StrictLogging {
     val connections = conf.getProperty("cassandra.maxpoolsize").toInt
 
     logger.info("Connecting to Cassandra servers '{}'", seeds)
+    datacenter.foreach(logger.info("Configuring Cassandra client for datacenter-local in {}", _))
```
jacobsi: How about a message that always prints, but shows that the value is unset, when it is unset, and outputs the actual value in all other cases?

jacobsi:
lgtm

PR #119, opened 2016-06-17
EN-3361 Add write endpoint to balboa-http

cramertj:


balboa-http/src/main/scala/com/socrata/balboa/server/MainServlet.scala
```diff
@@ -0,0 +1,79 @@
...
+  // Match paths like /metrics/:entityId and /metrics/:entityId/whatever
+  get("""^\/metrics\/([^\/]*).*""".r) {
+    val entityId = params("captures")
+    val response = MetricsRest.get(entityId, params, getAccepts(request))
+    contentType = response.contentType
+    response.result
+  }
+
+  post("/metrics/:entityId") {
+    // Note: content type must be set -before- calling extractOpt for the body to be sent as JSON
```
jacobsi: Can you expand on this a little? It's hard to me to understand how getting the parsed body of the request affects the response. What happens?

balboa-http/src/main/scala/com/socrata/balboa/server/EntityJSON.scala
```diff
@@ -0,0 +1,5 @@
+package com.socrata.balboa.server
+
+case class EntityJSON(timestamp: Long, metrics: Map[String, MetricJSON])
```
jacobsi: JSON isn't quite accurate in the name. This is a class, not JSON. If it sometimes get serialized to JSON that's just a representation detail.

Isn't this class almost the same as the `Metrics` object? Why is there a new class?

balboa-http/src/main/scala/com/socrata/balboa/server/Main.scala
```diff
@@ -1,48 +1,14 @@
...
-import com.socrata.balboa.metrics.config.Configuration
-import com.socrata.balboa.server.rest.{EntitiesRest, MetricsRest, VersionRest}
-import com.socrata.http.routing.{HttpMethods, SimpleRoute, SimpleRouter}
-import com.socrata.http.server.implicits._
-import com.socrata.http.server.responses._
-import com.socrata.http.server.{Filter, HttpResponse, Service, SimpleFilter, SocrataServerJetty}
-import com.typesafe.scalalogging.slf4j.StrictLogging
-import org.apache.log4j.PropertyConfigurator
+object Main extends App {
+  val logger = LoggerFactory.getLogger(getClass)
```
jacobsi: Please use `with StrictLogging` instead of putting an explicit member here.

balboa-http/src/main/scala/com/socrata/balboa/server/MainServlet.scala
cramertj: It fails to set the response type. I'd imagine it's an HTTP implementation detail having to do with reading in the request body before setting headers.

balboa-http/src/main/scala/com/socrata/balboa/server/EntityJSON.scala
cramertj: This case class represents precisely the entity JSON encoded in the request body of the POST request. This allows Jackson / Json4s to convert directly to and from JSON with out any extra marshalling methods, similar to how Rojoma Json classes were being used in the old balboa-http. Also note that it holds a timestamp, something not contained in the Metrics object.

jacobsi: It would be better to do this by updating the existing objects instead of creating new ones, but as discussed offline, this will be consolidated in a following review.

balboa-http/src/main/scala/com/socrata/balboa/server/rest/MetricsRest.scala
```diff
@@ -32,10 +34,9 @@ object MetricsRest {
   val periodMeter = com.yammer.metrics.Metrics.newTimer(
     classOf[MetricsRest], "period queries", TimeUnit.MILLISECONDS, TimeUnit.SECONDS)
 
-  val ds = DataStoreFactory.get()
 
   def get(entityId: String, params: Params, accepts: Seq[String]): ResponseWithType = {
-    val period = params.get(PeriodKey).map(Extractable[Period].extract) match {
+    val period = params.get("period").map(Extractable[Period].extract) match {
```
jacobsi: Didn't you just introduce the `PeriodKey` constant in a recent commit? Are you taking it out here?

cramertj: That was a mistake in the merge. Reverting it.

balboa-http/src/main/scala/com/socrata/balboa/server/rest/MetricsRest.scala
```diff
@@ -67,6 +68,21 @@ object MetricsRest {
...
+    for ((name, metric) <- entity.metrics) {
+      val recordType = metric.`type` match {
+        case "ABSOLUTE" => RecordType.ABSOLUTE
+        case "AGGREGATE" => RecordType.AGGREGATE
+        case _ => return badRequest("metric type", "must be ABSOLUTE or AGGREGATE")
+      }
+      metrics.put(name, new Metric(recordType, metric.value))
+    }
+    dataStore.persist(entityId, entity.timestamp, metrics)
+    ResponseWithType(json, Ok("{}"))
```
jacobsi: I think this should return 204 and an empty body.

jacobsi:
lgtm

PR #120, opened 2016-06-17
EN-3408 Update Balboa Agent to write to new HTTP API on balboa-http

cramertj:
No tests yet, but I'm opening the PR for review.


balboa-agent/src/main/java/com/socrata/balboa/agent/HttpOrMq.java
```diff
@@ -0,0 +1,19 @@
+package com.socrata.balboa.agent;
+
+public enum HttpOrMq {
```
jacobsi: `TransportType` would be a better name. Supports the expansion, or contraction, of the enumeration in the future.

jacobsi: Why is this a Java class? Socrata is transitioning to Scala, so new code should be in Scala if at all possible.

balboa-agent/src/main/scala/HttpMetricQueue.scala
```diff
@@ -0,0 +1,61 @@
...
+   * @param timestamp Time when this metric was created.
+   * @param recordType Type of metric to add, See [[Metric.RecordType]] for more information.
+   */
+  override def create(entity: IdParts,
+             name: IdParts,
+             value: Long,
+             timestamp: Long = new Date().getTime,
+             recordType: Metric.RecordType = Metric.RecordType.AGGREGATE): Unit = {
+
+    // TODO: lazy load metrics for the same entity and send them over in batches
```
jacobsi: What is this for?

balboa-agent/src/main/scala/com/socrata/balboa/agent/BalboaAgent.scala
```diff
@@ -34,81 +37,19 @@ object BalboaAgent extends App with Config with StrictLogging {
...
-  } catch {
-    // Known causes of an exception here are poorly formatted connection string.
-    // No amount of retry will fix that, so just exit.
-    case e: Throwable =>
-      logger error("Unable to initialize ActiveMQ connection.", e)
-      sys.exit(1)
+  val metricQueue = reportStrategy match {
+    case HttpOrMq.MQ => amqMetricQueue()
+    case HttpOrMq.HTTP =>
+      //throw new RuntimeException("Metric reporting by HTTP not yet supported")
```
jacobsi: Can this line be deleted now?

balboa-agent/src/main/scala/HttpMetricQueue.scala
```diff
@@ -0,0 +1,61 @@
...
+
+    val request = POST(url).setBody(pretty(render(Extraction.decompose(metricToWrite)))).apply
+
+    val requestWithTimeout = Future { Await.result(request, timeout) }
+
+    requestWithTimeout.onComplete {
+      case Success(response) =>
+        val responseCode = response.code.code
+        if (responseCode != Ok.code) {
+          logger info s"HTTP POST to Balboa HTTP returned error code $responseCode: " + response.bodyString
```
jacobsi: If this is a Scala string, then the `response.bodyString` can be embedded too.

balboa-agent/src/main/scala/HttpMetricQueue.scala
```diff
@@ -0,0 +1,61 @@
...
+        }
+      case Failure(failure) =>
+        logger info s"HTTP POST to Balboa HTTP failed with " + failure.getMessage
```
jacobsi: A Scala string with `failure.getMessage` embedded, or it doesn't need to be a Scala string.

balboa-agent/src/main/scala/com/socrata/balboa/agent/Config.scala
```diff
@@ -29,6 +36,18 @@ trait Config extends JMSClientConfig {
...
   def initialDelay(defaultDelay: Long = 0): Long = Configuration.get().getLong(Keys.InitialDelay, defaultDelay)
 
+  def httpOrMQ(defaultMethod: HttpOrMq = HttpOrMq.HTTP): HttpOrMq =
+    Configuration.get().getString(Keys.HttpOrMq, defaultMethod.toString) match {
+      case "HTTP" => HttpOrMq.HTTP
+      case "MQ" => HttpOrMq.MQ
+      case _ => defaultMethod
+    }
+
+  def balboaHttpUrl(defaultUrl: String): String =
```
jacobsi: It looks like you added this method, but the only call site I can find has an empty string as the default value. That pretty much means there doesn't need to be a default value parameter.

What about returning an Option, and if None or empty, log an error and exit. I can't think of a way to make it a recoverable scenario if there is no balboa-http endpoint specified.

balboa-agent/src/main/java/com/socrata/balboa/agent/HttpOrMq.java
cramertj: It's a Java enum because plain Scala enums are really annoying to work with. If you'd prefer, I can write it as a sealed trait.

balboa-agent/src/main/scala/HttpMetricQueue.scala
cramertj: If there are a lot of metrics being sent with the same entity, they can be send over in the same POST request rather than each one getting its own HTTP request. It would cut down on overhead, but it's not necessary for functionality.

balboa-agent/src/main/scala/com/socrata/balboa/agent/BalboaAgent.scala
cramertj: Yup!

balboa-agent/src/main/scala/HttpMetricQueue.scala
cramertj: Sure. I thought `+ response.bodyString` was maybe cleaner than `${response.bodyString}"`, but I'll change it.

balboa-agent/src/main/java/com/socrata/balboa/agent/HttpOrMq.java
jacobsi: Yes, Scala please.

balboa-agent/src/main/scala/HttpMetricQueue.scala
cramertj: Oh, I probably made it embedded and then took it out but left the `s`.

balboa-agent/src/main/scala/HttpMetricQueue.scala
jacobsi: Could you make a Jira describing the update. Assign it to the admin team, then delete this TODO. TODOs are code litter. Unacknowledged technical debt.

balboa-agent/src/main/scala/com/socrata/balboa/agent/Config.scala
cramertj: Yeah, I wasn't sure what the future use cases might be for this API, so I decided to just mirror the behavior of the other conf getters. I will change it to return an option.

balboa-agent/src/main/scala/HttpMetricQueue.scala
cramertj: Sure, I just figured I'd do it before this got merged in. But that works too!

balboa-agent/src/main/scala/com/socrata/balboa/agent/Config.scala
```diff
@@ -29,6 +36,18 @@ trait Config extends JMSClientConfig {
 
   def initialDelay(defaultDelay: Long = 0): Long = Configuration.get().getLong(Keys.InitialDelay, defaultDelay)
 
+  def httpOrMQ(defaultMethod: TransportType = Mq): TransportType =
```
jacobsi: So when the enum name was changed, that should cascade to all the other places it was used. Like this method name, and the config file.

cramertj: Agreed. Missed that one.

balboa-agent/src/main/scala/com/socrata/balboa/agent/BalboaAgent.scala
```diff
@@ -155,4 +97,82 @@ object BalboaAgent extends App with Config with StrictLogging {
...
+      "(This is setting the username, password and destination.)"
+    val amqConnectionFactory = (activemqUser, activemqPassword) match {
+      case (Some(user), Some(password)) => new ActiveMQConnectionFactory(user, password, activemqServer)
+      case _                            => new ActiveMQConnectionFactory(activemqServer)
+    }
+
+    // The ActiveMQ libraries can be obscure when they are misbehaving. The hope
+    // is that by registering this listener, additional information will be
+    // written to the logs for troubleshooting.
+    amqConnectionFactory.setTransportListener(new TransportListener {
```
jacobsi: You can shorten this method by moving the TransportListener to it's own concrete class somewhere.

If that isn't enough to get rid of the warning, I see two try blocks down below that can be refactored to methods.

balboa-agent/src/main/scala/com/socrata/balboa/agent/BalboaAgent.scala
```diff
@@ -108,24 +108,7 @@ object BalboaAgent extends App with Config with StrictLogging {
...
-      // as well. So this log message will appear after the call to `.start` down
-      // below.
-      override def transportInterupted(): Unit =
-        logger warn "ActiveMQ connection is being closed or has suffered an interruption " +
-          "from which it hopes to recover."
-
-      override def transportResumed(): Unit =
-        logger warn "ActiveMQ connection is starting up or has resumed after an interruption."
-    })
+    amqConnectionFactory.setTransportListener(new LoggingTransportListener())
```
jacobsi: Is this enough to get rid of the `// scalastyle:off` annotations?

cramertj: No, the method is still longer than 50 lines.

balboa-agent/src/main/scala/HttpMetricQueue.scala
```diff
@@ -0,0 +1,60 @@
...
+
+    val metricToWrite = EntityJSON(timestamp, Map(name.toString -> MetricJSON(value, recordType.toString)))
+
+    val url = new URL(s"$balboaHttpURL/metrics/$entity")
+
+    val request = POST(url).setBody(pretty(render(Extraction.decompose(metricToWrite)))).apply
+
+    val requestWithTimeout = Future { Await.result(request, timeout) }
+
+    requestWithTimeout.onComplete {
```
jacobsi: I think the approach to take here is to just keep looping, infinitely and retrying when the send fails. Failed requests will wait, then retry. You can do a little exponential backoff up to a maximum wait time of lets say 2 minutes (stored in the config file) and then keep retrying every 2 minutes.

It's naive, and we may have to tune it a little bit before releasing but my reasoning (such as it is) goes thusly.

The previous behavior was infinite loop on a failed send. If we change that behavior (and propagate exceptions), we will have to change behavior up in the call stack, but the ActiveMQ connection would still behave the same. Probably more work than we need at this time.

Which reminds me, we should be logging improperly formatted metric save requests. Not sure what's happening right now. I added a Jira for you to confirm/add the behavior.

balboa-agent/src/main/resources/config/config.properties
```diff
@@ -4,6 +4,10 @@
 # Current configuration for the Metric Consumer.
 balboa.agent.data.dir = /tmp
 balboa.agent.sleeptime = 10000
+balboa.agent.transport.type = MQ
+# balboa.agent.balboa.http.url
+# balboa.agent.balboa.http.timeout
+# balboa.agent.balboa.http.max.retry.wait
```
jacobsi: Are these here as examples?

balboa-agent/src/main/scala/com/socrata/balboa/agent/BalboaAgent.scala
```diff
@@ -137,4 +97,65 @@ object BalboaAgent extends App with Config with StrictLogging {
     }
   })
 
+
+  // scalastyle:off method.length
+  def amqMetricQueue(): MetricQueue = {
+    logger info "Initializing ActiveMQ connection factory. " +
```
jacobsi: An analogous comment for the Http method would be good.

balboa-agent/src/main/resources/config/config.properties
cramertj: I put them there so that people would know what keys were available.

jacobsi: How about putting in some usable default values:

    balboa.agent.balboa.http.url = "http://balboa.app.aws-us-west-2-staging.socrata.net/"
    balboa.agent.balboa.http.timeout = 5
    balboa.agent.balboa.http.max.retry.wait = 120

jacobsi: Can you add `.seconds` to the configuration item names so the units are visible to the reader?

balboa-agent/src/main/scala/com/socrata/balboa/agent/BalboaAgent.scala
```diff
@@ -32,65 +35,22 @@ object BalboaAgent extends App with Config with StrictLogging {
   logger info "Starting the JMX Reporter."
   jmxReporter.start()
 
-  logger info "Initializing ActiveMQ connection factory. (This is setting the username, password and destination.)"
-  val amqConnectionFactory = (activemqUser, activemqPassword) match {
-    case (Some(user), Some(password)) => new ActiveMQConnectionFactory(user, password, activemqServer)
-    case _                            => new ActiveMQConnectionFactory(activemqServer)
-  }
+  private val reportStrategy = this.transportType()
```
jacobsi: `transportType` would be a better name than `reportStrategy`.

balboa-agent/src/main/scala/com/socrata/balboa/agent/BalboaAgent.scala
```diff
@@ -32,65 +35,22 @@ object BalboaAgent extends App with Config with StrictLogging {
...
-  // logic to reconnecting to a server when the connection is lost.
-  val amqConnection = try {
-    amqConnectionFactory.createConnection().asInstanceOf[ActiveMQConnection]
-  } catch {
-    // Known causes of an exception here are poorly formatted connection string.
-    // No amount of retry will fix that, so just exit.
-    case e: Throwable =>
-      logger error("Unable to initialize ActiveMQ connection.", e)
-      sys.exit(1)
+  val metricQueue = reportStrategy match {
```
jacobsi: Could you rename this from `metricQueue` to `metricPublisher`"

balboa-agent/src/main/scala/com/socrata/balboa/agent/HttpMetricQueue.scala
```diff
@@ -0,0 +1,72 @@
...
+             recordType: Metric.RecordType = Metric.RecordType.AGGREGATE): Unit = {
+
+    val metricToWrite = EntityJSON(timestamp, Map(name.toString -> MetricJSON(value, recordType.toString)))
+    val url = new URL(s"$balboaHttpURL/metrics/${URLEncoder.encode(entity.toString, Utf8)}")
+    val request = POST(url).setBody(pretty(render(Extraction.decompose(metricToWrite))))
+
+    var waitInLoop = StartingWaitDuration
+
+    while (true) {
+      val requestWithTimeout = Future { Await.result(request.apply, timeout) }
```
jacobsi: This seems odd to me. The purpose of Await is to serialize a Future. Which means `request.apply` is returning a Future. But you are wrapping it in another future.

Since `requestWithTimeout` is a future, that means it is executing in the background, and the loop will wait, come back here and start another `request.apply` in the background. So if the server is slow, there is a race condition where multiple requests can be submitted.

jacobsi: How did you test this retry stuff?

balboa-agent/src/main/resources/config/config.properties
```diff
@@ -4,6 +4,12 @@
 # Current configuration for the Metric Consumer.
 balboa.agent.data.dir = /tmp
 balboa.agent.sleeptime = 10000
+balboa.agent.transport.type = MQ
+
+# Example properties for transport.type = HTTP
+balboa.agent.balboa.http.url = "http://balboa.app.aws-us-west-2-staging.socrata.net/"
+balboa.agent.balboa.http.timeout.ms = 150
+balboa.agent.balboa.http.max.retry.wait.ms = 2000
```
jacobsi: You can leave them commented out, as long as they are filled in.

balboa-agent/src/main/resources/config/config.properties
```diff
@@ -4,6 +4,12 @@
 # Current configuration for the Metric Consumer.
 balboa.agent.data.dir = /tmp
 balboa.agent.sleeptime = 10000
+balboa.agent.transport.type = MQ
```
jacobsi: Could you comment this one out and put a comment above it indicating the two possible values and what they mean?

jacobsi: Never mind. Did this.

balboa-agent/src/main/resources/config/config.properties
jacobsi: Did this too.

balboa-agent/src/main/scala/com/socrata/balboa/agent/HttpMetricQueue.scala
```diff
@@ -0,0 +1,87 @@
...
+    // An improvement on this naive infinite retry policy could go 1 of two ways:
+    // 1. Adapt `balboa-http` to accept all transmissions and persist the ones
+    //    it can't parse, retrying them later (when it's been upgraded for
+    //    example). This would safer for keeping accurate metrics.
+    // 2. Make `balboa-agent` respond differently to different HTTP error
+    //    codes. Possibly, for example (but not exhaustively) 404, 500s are
+    //    infinite retry, but 400s are written to a file to be retried in the
+    //    future.
+    while (true) {
+      val requestWithTimeout = Future { Await.result(request.apply, timeout) }
```
jacobsi: This seems odd to me. The purpose of `Await` is to serialize a `Future`. Which means `request.apply` is returning a `Future`. But you are wrapping `Await` back into another `Future`.

Since `requestWithTimeout` is a `Future`, that means it is executing in the background, and the loop will wait, come back here and start another `request.apply` in the background. So if the server is slow, there is a race condition where multiple requests can be submitted.

jacobsi: How did you test this?

cramertj:
@jacobsi Any ideas on the best approach to writing tests for this one? Should I copy the approach of the IT tests in bal-http and write a bash script to stand up the http server and then try and persist data? Is there another approach that makes more sense?

balboa-agent/src/main/scala/com/socrata/balboa/agent/BalboaAgent.scala
jacobsi: Excuse me, I meant an analogous log line.

jacobsi:
I've made a change to `HttpMetricQueue` that will make it easier to test. You should be able to create some unit tests to cover it now by passing in a mock `HttpClient`. You should also look through the newman documentation to see if there is an implementation of the HttpClient that is suitable for use in testing. It might be a mock of some kind, or a test implementation.

balboa-agent/src/test/scala/com/socrata/balboa/agent/HttpMetricQueueSpec.scala
```diff
@@ -26,12 +29,32 @@ class HttpMetricQueueSpec extends WordSpec
...
+  implicit def convertJSONAssertion(j: => String): AssertionJSON = new AssertionJSON(j)
+
+
   implicit val execContext = scala.concurrent.ExecutionContext.global
 
   val TestUrl = "http://www.google.com"
   val Timeout = 20.millis
   val MaxRetryWait = 500.millis
 
+  val TestTimeout = 5000 // millis
```
jacobsi: Why not use `.millis.toMillis`? Comments aren't checked by the compiler.

cramertj: Sure, that works. It just seems kind of odd.

jacobsi:
lgtm

PR #121, opened 2016-06-21
EN-2828 Upgrade Balboa Cassandra Driver to Datastax

cramertj:


balboa-core/src/main/scala/com/socrata/balboa/metrics/data/impl/Cassandra11QueryImpl.scala
```diff
@@ -1,43 +1,49 @@
...
+
+    /*
     // aggregate column family
     val aggResults: ColumnList[String] = fetch_cf(RecordType.AGGREGATE, entityKey, period).getResult
     aggResults.asScala.map(c => ret.put(c.getName, new Metric(Metric.RecordType.AGGREGATE, c.getLongValue())))
 
     // absolutes column family
     val absResults: ColumnList[String] = fetch_cf(RecordType.ABSOLUTE, entityKey, period).getResult
     absResults.asScala.map(c => ret.put(c.getName, new Metric(Metric.RecordType.ABSOLUTE, c.getLongValue())))
+    */
```
jacobsi: Please delete. We can find this with git history.

balboa-core/src/main/scala/com/socrata/balboa/metrics/data/impl/Cassandra11QueryImpl.scala
```diff
@@ -46,12 +52,24 @@ class Cassandra11QueryImpl(context: AstyanaxContext[Keyspace]) extends Cassandra
...
+      val retVal = asScalaIterator(rows.iterator()).map(_.getString("key")).map(removeTimestamp)
+
+      /*
       val retVal: Iterator[String] = context.getEntity.prepareQuery(Cassandra11Util.getColumnFamily(period, recordType))
         .setConsistencyLevel(ConsistencyLevel.CL_ONE)
         .withRetryPolicy(new ExponentialBackoff(250, 5)) // initial, max tries
         .getAllRows
         .setRowLimit(100) // max 100 rows per query to cassandra
         .execute().getResult.iterator.asScala.map(removeTimestamp)
+      */
```
jacobsi: Please delete.

balboa-core/src/main/scala/com/socrata/balboa/metrics/data/impl/Cassandra11QueryImpl.scala
```diff
@@ -87,38 +118,78 @@ class Cassandra11QueryImpl(context: AstyanaxContext[Keyspace]) extends Cassandra
...
 
+    val batchStatement = new BatchStatement(BatchStatement.Type.LOGGED)
+
+    val entityKeyWhere = QueryBuilder.eq("key", entityKey)
 
+    /*
     val m:MutationBatch = context.getEntity.prepareMutationBatch
       .setConsistencyLevel(ConsistencyLevel.CL_ONE)
       .withRetryPolicy(new ExponentialBackoff(250, 5))
+    */
```
jacobsi: Please delete.

balboa-http/src/main/resources/config/config.properties
```diff
@@ -26,8 +26,8 @@ balboa.datastore = cassandra
 memcached.servers = localhost:11211
 
 # Cassandra servers. Servers should be separated by a comma.
-cassandra.servers = localhost:9160
-cassandra.keyspace = Metrics2012
+cassandra.servers = localhost:9042
+cassandra.keyspace = "Metrics2012"
```
jacobsi: Could you stick a comment in here that explains the necessity of this? Keeping in mind that not only the future person modifying this code will read it, but also other people responsible for converting from Astyanax drivers to Datastax drivers will see it and need to apply this to their code base.

project/Dependencies.scala
```diff
@@ -13,6 +13,7 @@ object Dependencies {
   private object versions {
     val astyanax = "1.56.26"
```
jacobsi: Can this go?

balboa-core/src/main/scala/com/socrata/balboa/metrics/data/impl/Cassandra11Util.scala
```diff
@@ -109,28 +108,37 @@ object Cassandra11Util extends StrictLogging {
...
-      .forKeyspace(keyspace)
-      .withAstyanaxConfiguration(new AstyanaxConfigurationImpl()
-        .setDiscoveryType (NodeDiscoveryType.RING_DESCRIBE)
-      )
-      .withConnectionPoolConfiguration(connectionPoolConfiguration)
-      .withConnectionPoolMonitor(new CountingConnectionPoolMonitor())
-      .buildKeyspace(ThriftFamilyFactory.getInstance())
-    cxt.start()
-    cxt
+    //datacenter.foreach(connectionPoolConfiguration.setLocalDatacenter)
```
jacobsi: Can delete this line.

balboa-core/src/main/scala/com/socrata/balboa/metrics/data/impl/Cassandra11QueryImpl.scala
```diff
@@ -88,30 +93,39 @@ class Cassandra11QueryImpl(context: AstyanaxContext[Keyspace]) extends Cassandra
...
-    if (absolutes.nonEmpty) {
-      var cols = m.withRow(Cassandra11Util.getColumnFamily(period, RecordType.ABSOLUTE), entityKey)
-      for { (k,v) <- absolutes } {
-        if (k != ""){ cols = cols.putColumn(k, v.getValue.longValue) }
-        else { logger warn("dropping metric with empty string as column") }
+    val batchStatement = new BatchStatement(BatchStatement.Type.LOGGED)
+
+    val entityKeyWhere = QueryBuilder.eq("key", entityKey)
+
+    for (sameTypeRecords <- List(aggregates, absolutes)) {
```
jacobsi: How about

    for (sameTypeRecords <- List(aggregates, absolutes).filter(_.nonEmpty)) {

and removing a layer of nesting from the loop. Would that work?

balboa-core/src/main/scala/com/socrata/balboa/metrics/data/impl/Cassandra11Util.scala
```diff
@@ -109,28 +108,36 @@ object Cassandra11Util extends StrictLogging {
...
-      .withConnectionPoolMonitor(new CountingConnectionPoolMonitor())
-      .buildKeyspace(ThriftFamilyFactory.getInstance())
-    cxt.start()
-    cxt
+
+    val dcPolicy = DCAwareRoundRobinPolicy.builder()
+    datacenter.foreach(dc => dcPolicy.withLocalDc(dc))
+
+    val poolingOptions = new PoolingOptions()
+      .setIdleTimeoutSeconds(sotimeout)
```
jacobsi: Don't set this. The default should be fine. In any case, this is not the right value. A connection idle timeout is on the order of 100s of seconds. A connect timeout (how long to timeout while the TCP connection is establishing) is on the order of seconds. These are very different numbers.

balboa-core/src/main/scala/com/socrata/balboa/metrics/data/impl/Cassandra11Util.scala
```diff
@@ -109,28 +108,36 @@ object Cassandra11Util extends StrictLogging {
...
+      .setPoolTimeoutMillis(sotimeout)
```
jacobsi: The order of magnitude is right here, but I don't think something called `sotimeout` = `socket timeout` is the right name for the property in the properties file. Default is probably okay here, so you can remove this line too.

balboa-core/src/main/scala/com/socrata/balboa/metrics/data/impl/Cassandra11Util.scala
```diff
@@ -109,28 +108,36 @@ object Cassandra11Util extends StrictLogging {
...
+      .setMaxConnectionsPerHost(HostDistance.REMOTE, connections)
+
+    val seedInetAddrs = seeds.split(",").map(addrStr => {
+      val addrAndPort = addrStr.split(":")
+      if (addrAndPort.length != 2) {
+        throw new IllegalArgumentException("Address and port must be separated by a ':' in: " + addrStr)
+      }
+      new InetSocketAddress(addrAndPort(0), addrAndPort(1).toInt)
+    })
+
```
jacobsi: A log statement here that list the Cassandra IP:port endpoints that are configured is nice.

balboa-core/src/main/scala/com/socrata/balboa/metrics/data/impl/Cassandra11Util.scala
```diff
@@ -109,28 +108,36 @@ object Cassandra11Util extends StrictLogging {
...
+    datacenter.foreach(dc => dcPolicy.withLocalDc(dc))
+
+    val poolingOptions = new PoolingOptions()
+      .setIdleTimeoutSeconds(sotimeout)
+      .setPoolTimeoutMillis(sotimeout)
+      .setMaxConnectionsPerHost(HostDistance.LOCAL, connections)
+      .setMaxConnectionsPerHost(HostDistance.REMOTE, connections)
+
+    val seedInetAddrs = seeds.split(",").map(addrStr => {
+      val addrAndPort = addrStr.split(":")
```
jacobsi: `.trim()` before the `.split(...)` would be a nice addition.

balboa-core/src/main/scala/com/socrata/balboa/metrics/data/impl/Cassandra11QueryImpl.scala
```diff
@@ -1,41 +1,39 @@
 package com.socrata.balboa.metrics.data.impl
 
 import java.io.IOException
+import java.nio.charset.Charset
```
jacobsi: Is this needed? I don't see it used anywhere.

balboa-core/src/main/scala/com/socrata/balboa/metrics/data/impl/Cassandra11QueryImpl.scala
```diff
@@ -44,12 +42,15 @@ class Cassandra11QueryImpl(context: AstyanaxContext[Keyspace]) extends Cassandra
   def getAllEntityIds(recordType: RecordType, period: Period): Iterator[String] = {
     fastfail.proceedOrThrow()
     try {
-      val retVal: Iterator[String] = context.getEntity.prepareQuery(Cassandra11Util.getColumnFamily(period, recordType))
-        .setConsistencyLevel(ConsistencyLevel.CL_ONE)
-        .withRetryPolicy(new ExponentialBackoff(250, 5)) // initial, max tries
```
jacobsi: It looks like the retry policy isn't part of the query execution any more, however it seems to be set when establishing the cluster. The default seems to be the best choice, so that should be fine. However it would be nice to wrap that in the logging retry policy. If I've understood correctly, that should just add some information to the logs about what is happening.

balboa-core/src/main/scala/com/socrata/balboa/metrics/data/impl/Cassandra11QueryImpl.scala
```diff
@@ -1,41 +1,39 @@
...
+import com.datastax.driver.core.{BatchStatement, ConsistencyLevel, Row}
 import com.socrata.balboa.metrics.Metric.RecordType
+import com.socrata.balboa.metrics.data.impl.Cassandra11Util.DatastaxContext
 import com.socrata.balboa.metrics.data.{BalboaFastFailCheck, Period}
 import com.socrata.balboa.metrics.{Metric, Metrics}
 import com.typesafe.scalalogging.slf4j.StrictLogging
 
-import scala.collection.JavaConverters._
 import scala.{collection => sc}
+import scala.collection.JavaConversions.asScalaIterator
```
jacobsi: Apparently there are problems with JavaConversions. `Conversions are perversions` is the matra. Prefer `JavaConverters`.

balboa-core/src/main/scala/com/socrata/balboa/metrics/data/impl/Cassandra11QueryImpl.scala
cramertj: I'm going to set the LoggingPolicy on as the RetryPolicy in Cassandra11Util.scala, rather than here (as it would only affect that specific query).

balboa-core/src/main/scala/com/socrata/balboa/metrics/data/impl/Cassandra11Util.scala
```diff
@@ -109,28 +108,41 @@ object Cassandra11Util extends StrictLogging {
...
+
+    val seedInetAddrs = seeds.split(",").map(addrStr => {
+      val addrAndPort = addrStr.trim().split(":")
+      if (addrAndPort.length != 2) {
+        throw new IllegalArgumentException("Address and port must be separated by a ':' in: " + addrStr)
+      }
+      new InetSocketAddress(addrAndPort(0), addrAndPort(1).toInt)
+    })
+
+    logger info s"Connecting to Cassandra on seed addresses: ${ seedInetAddrs.fold("")(_ + ", " + _) }"
```
jacobsi: It doesn't seem like this will produce the right output. Won't there will be a leading `,`. Why not use mkString?

    seedInetAddrs.mkString(", ")

balboa-core/src/main/scala/com/socrata/balboa/metrics/data/impl/Cassandra11QueryImpl.scala
jacobsi: Sounds good.

jacobsi:
lgtm

PR #123, opened 2016-06-23
Split balboa HTTP into servlets

cramertj:


balboa-http/src/main/scala/com/socrata/balboa/server/MetricsServlet.scala
```diff
@@ -26,31 +30,48 @@ object MetricsRest {
...
+    val entityId = params("captures")
 
-  def get(entityId: String, params: Params, accepts: Seq[String]): ResponseWithType = {
     val period = params.get(PeriodKey).map(Extractable[Period].extract) match {
       case Some(Right(value)) => value
-      case Some(Left(err)) => return badRequest(PeriodKey, err)
-      case None => return required(PeriodKey)
+      case Some(Left(err)) =>
+        contentType = json
+        return badRequest(PeriodKey, err).result
```
cramertj: @jacobsi These `.result` field accesses, as well as the `contentType = x` will all disappear when we remove the Protobuf API.

jacobsi:
lgtm

PR #125, opened 2016-06-24
Quote mixed case keyspace names.

jacobsi:
Cassandra does not handle mixed case names well, and it would be better
if people never used mixed case names.

If the keyspace name is mixed case, quote it before passing it to the
Datastax driver.


raganw:
Looks good to me.  Merging so as to get it into staging, as staging is still not happy (we rolled back to a version of the code that didn't like the updated configuration).

PR #126, opened 2016-06-24
Renamed Cassandra related classes.

jacobsi:
It appears that Long Long Ago, someone named some Cassandra related classes with `11`, which I believe indicated Cassandra 1.1. That is no longer true, and will get less true as time goes on.


housepage:
:ship it:

PR #128, opened 2016-06-24
EN7345 no more yammer

cramertj:


balboa-http/src/main/scala/com/socrata/balboa/server/MetricsServlet.scala
```diff
@@ -29,15 +29,6 @@ class MetricsServlet extends JacksonJsonServlet
   val CombineKey = "combine"
   val FieldKey = "field"
 
-  val seriesMeter = com.yammer.metrics.Metrics.newTimer(
```
cramertj: @jacobsi It doesn't look like these were being used or displayed anywhere, so I've just removed them. It might be worth investigating a better long-term replacement along the lines of EN-679.

balboa-http/src/main/scala/com/socrata/balboa/server/MetricsServlet.scala
```diff
@@ -85,10 +76,6 @@ class MetricsServlet extends JacksonJsonServlet
       contentType = response.contentType
       response.result
     }
-    finally
-    {
-      periodMeter.update(System.currentTimeMillis() - begin, TimeUnit.MILLISECONDS)
```
jacobsi: This looks like a timer metric to me. You can update it to the style described here:

http://scalatra.org/2.4/guides/monitoring/metrics.html#toc_214

Same with the other two I think.

cramertj: I looked at the link and searched around. I can't seem to find an example of someone using a timer in the way we were using the yammer one, that is to say, calling an update method with a provided time. Adding the MetricsSupport trait gives you the ability to specify a block that gets run on a timer: ```timer("timer") { // code here }```, but that's not quite what we want. I'm going to check into the dropwizard bits directly seeing as they're already used elsewhere.

jacobsi: Why don't you think a timer block is what we want?

cramertj: What do you envision going in the timer block? I thought the idea was to simply update a timer every time a request is processed (e.g. period-related timer gets updated every time a period request completes). I don't understand how that translates into a single callable block. The callable blocks are (to my understanding) recording the duration that it takes to run a Callable. From the codahale Timer:
```java
    /**
     * Adds a recorded duration.
     *
     * @param duration the length of the duration
     * @param unit     the scale unit of {@code duration}
     */
    public void update(long duration, TimeUnit unit) {
        update(unit.toNanos(duration));
    }

    /**
     * Times and records the duration of event.
     *
     * @param event a {@link Callable} whose {@link Callable#call()} method implements a process
     *              whose duration should be timed
     * @param <T>   the type of the value returned by {@code event}
     * @return the value returned by {@code event}
     * @throws Exception if {@code event} throws an {@link Exception}
     */
    public <T> T time(Callable<T> event) throws Exception {
        final long startTime = clock.getTick();
        try {
            return event.call();
        } finally {
            update(clock.getTick() - startTime);
        }
    }
```
It looks like the timer block corresponds to the "time" function, when we're trying to use the "update" function.

jacobsi: Have a look at the example I added to my comment above and look at the code we are replacing. I'll chat with you about this a bit, I have a meeting now.

cramertj: I believe I understand the source of my confusion-- I misunderstood what was going on in the syntax of the scalatra timer blocks (I didn't realize they were running code in-place rather than on a callback like the `get` blocks. It made sense when I remembered that {} is just the syntax for passing a closure that takes no arguments, and that it's not some weird method declaration syntax or anything more complicated). I should be able to sort it out pretty easily now.

balboa-http/src/main/scala/com/socrata/balboa/server/MetricsServlet.scala
```diff
@@ -67,28 +59,23 @@ class MetricsServlet extends JacksonJsonServlet
...
       contentType = json
       return malformedDate(date).result
     })
 
-    try
-    {
-      val iter = dataStore.find(entityId, period, range.start, range.end)
-      var metrics = Metrics.summarize(iter)
+    timer("period queries") {
+      try {
```
jacobsi: I don't think you need this try block any more, do you? No catch clause, no finally.

balboa-http/src/main/scala/com/socrata/balboa/server/MetricsServlet.scala
```diff
@@ -121,23 +108,18 @@ class MetricsServlet extends JacksonJsonServlet
       return unacceptable
     })
 
-    val begin = System.currentTimeMillis()
+    timer("range queries") {
+      try {
```
jacobsi: Same here, I don't think the try clause is needed.

balboa-http/src/main/scala/com/socrata/balboa/server/MetricsServlet.scala
```diff
@@ -177,18 +159,13 @@ class MetricsServlet extends JacksonJsonServlet
...
-      val body = renderJson(dataStore.slices(entityId, period, startDate, endDate)).getBytes(UTF_8)
-      val resp = ResponseWithType(json, Ok(body))
-      contentType = resp.contentType
-      resp.result
-    }
-    finally
-    {
-      seriesMeter.update(System.currentTimeMillis() - begin, TimeUnit.MILLISECONDS)
+    timer("series queries") {
+      try {
```
jacobsi: Here too.

balboa-http/src/main/scala/com/socrata/balboa/server/MetricsServlet.scala
```diff
@@ -69,15 +61,12 @@ class MetricsServlet extends JacksonJsonServlet
...
-    val begin = System.currentTimeMillis()
-
     val range = DateRange.create(period, ServiceUtils.parseDate(date).getOrElse {
       contentType = json
       return malformedDate(date).result
     })
 
-    try
-    {
+    timer("period queries") {
```
jacobsi: Why are you starting this here? It looks like previously the beginning time was above the range block?

cramertj: The timers are (I believe) intended to record how long it takes to query Cassandra and send back the results, not including the time necessary to parse the request. This is the pattern followed by the two other timer instances, and I believe we should keep consistent behavior between the timers.

Additionally, in the case that the date cannot be parsed, the API would return before querying Cassandra, causing timer entries far shorter than those that actually executed a query. Not only would this cause great variance in our timer, but it wouldn't match the behavior of the old system. In the old version, when a date fails to parse it returns immediately without updating the timer.

jacobsi: Sounds reasonable.

jacobsi:
lgtm

PR #129, opened 2016-06-24
EN6918 - Convert to Scala 2.11

cramertj:


jacobsi:
When I try this one out:

    sbt clean compile test
    ls -d */target/scala*/.

All I see is scala-2.10 builds. How did you test this one?

cramertj:
I was able to successfully compile even after bumping the build.sbt version, which I thought was the goal. Are we trying to do more than that? Perhaps I misunderstood the point of this ticket.

cramertj:
It's cross-compiling to 2.10 still, no? So you'd still expect to see the same output files as before, right?

jacobsi:
It doesn't cross compile on its own. It needs that command listed in the build.sbt file in order for cross-compiling to work.

You will know that it is correctly compiling in Scala 2.11 when the `sbt build` target produces scala-2.11 artifacts in the correct target directories. (ls -d */target/scala-*/.)

    sbt build

on its own isn't a test, since that has to work before and after the update.

To see the difference that cross compiling makes run this:

     sbt clean balboa-client-dispatcher/compile

Then look for the artifacts that it produces:

    ls -d balboa-client-dispatcher/target/scala-*/.

Then run a cross compile (indicated by the `+` in front of the sbt task):

    sbt "set crossScalaVersion = List("2.10.6", "2.11.7")" clean "+balboa-client-dispatcher/compile"

and run the same `ls` command as above to see the artifacts it produces.

project/BalboaAdmin.scala
```diff
@@ -5,6 +5,7 @@ import scoverage.ScoverageSbtPlugin
 
 object BalboaAdmin {
   lazy val settings: Seq[Setting[_]] = BuildSettings.projectSettings ++ Seq(
+    scalaVersion := "2.11.8",
```
jacobsi: Since many of these projects take their settings from BuildSettings, are you sure you need scalaVersion in each of these? It doesn't compile to scala 2.11 unless you put the setting in each sub project?

cramertj: I'm not sure. I didn't try changing buildSettings. I'll give that a shot.

cramertj: You were right! I could get rid of all of them except the buildSettings one.

project/Dependencies.scala
```diff
@@ -70,7 +70,8 @@ object Dependencies {
   val kafka_test = kafka % "test" classifier "test"
   val mockito_test = "org.mockito" % "mockito-core" % versions.mockito % "test"
   // Functionally Nice REST Client Currently newman only supports scala version 2.9.2 and 2.10
-  val newman = "com.stackmob" %% "newman" % versions.newman
+  // Because of this, its been replaced with megam newman, which supports scala 2.11
```
jacobsi: I wrote an improved comment here. It's gone now. If you are going to force push a change, you need to make sure you are up to date first.

jacobsi: And you shouldn't be force pushing changes in response to code reviews.

jacobsi:
lgtm

project/Dependencies.scala
cramertj: Yep. I realized that. Sorry.

cramertj: I've been force pushing so that I could rebase.

cramertj: But I'll stop doing that especially since squash merges make it entirely unnecessary

PR #131, opened 2016-06-27
ExecuteUpdate for DatastaxContext

cramertj:


balboa-core/src/main/scala/com/socrata/balboa/metrics/data/impl/CassandraUtil.scala
```diff
@@ -41,6 +41,12 @@ object CassandraUtil extends StrictLogging {
       session.close()
       result
     }
+
+    def executeUpdate(stmt: Statement): Unit = {
+      val session = this.newSession
+      session.execute(stmt)
+      session.close()
```
jacobsi: The `.close()` should be in the finally clause of a try block, so if there are any exceptions, it gets closed in any case. Same with the method above. We overlooked the error case last time.

cramertj: Good catch.

balboa-core/src/main/scala/com/socrata/balboa/metrics/data/impl/CassandraUtil.scala
```diff
@@ -37,9 +38,18 @@ object CassandraUtil extends StrictLogging {
 
     def execute(stmt: Statement): ju.List[Row] = {
       val session = this.newSession
-      val result = session.execute(stmt).all()
+      val result = Try(session.execute(stmt).all())
```
jacobsi: What happens to the exception here?

cramertj: It gets caught, then thrown on the call to `.get`. I can add a comment to that effect if you'd like. I considered doing it, but then wasn't sure if it would be obvious to people more used to scala than I.

jacobsi: Probably it's obvious to more experienced people.

jacobsi:
lgtm

PR #132, opened 2016-06-28
Fixed persisting of aggregate metrics

cramertj:
This fixes a bug when attempting to persist a combined collection of aggregate and non-aggregate metrics. Aggregate metrics must be persisted through a separate query from non-aggregates, as they require a counter batch, while non-aggregates require a non-counter batch. See discussion [here](http://www.datastax.com/dev/blog/atomic-batches-in-cassandra-1-2) and [here](https://datastax-oss.atlassian.net/browse/NODEJS-158).

### DIFF
--- a/balboa-http/src/it/scala/com/socrata/balboa/server/MetricsIntegrationTest.scala
+++ b/balboa-http/src/it/scala/com/socrata/balboa/server/MetricsIntegrationTest.scala
@@ -2,6 +2,7 @@ package com.socrata.balboa.server
 
 import java.net.URL
 
+import com.socrata.balboa.metrics.Metric.RecordType
 import com.socrata.balboa.metrics.data.DataStoreFactory
 import com.stackmob.newman.ApacheHttpClient
 import com.stackmob.newman.dsl.{GET, POST}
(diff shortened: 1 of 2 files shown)


balboa-http/src/it/scala/com/socrata/balboa/server/MetricsIntegrationTest.scala
```diff
@@ -219,19 +220,73 @@ class MetricsIntegrationTest extends FlatSpec with Matchers with BeforeAndAfterE
...
+  "Persist absolute metrics via POST" should "succeed" in {
+    persistManyMetrics(1 to 10, RecordType.ABSOLUTE)
+  }
+
+  "Persist aggregate metrics via POST" should "succeed" in {
+    persistManyMetrics(1 to 10, RecordType.AGGREGATE)
+  }
+
+  "Persist both absolute and aggregate metrics in the same request" should "succeed" in {
+    persistManyAbsoluteAndAggregate()
```
jacobsi: If you expanded `persistManyMetrics` to work like:

    persistManyMetrics(1 to 10, RecordType.ABSOLUTE)
    persistManyMetrics(1 to 10, RecordType.AGGREGATE)
    persistManyMetrics(1 to 10, RecordType.ABSOLUTE, RecordType.AGGREGATE)

You wouldn't need a second method.

cramertj: That's true. I was just kind of throwing stuff together quickly and figured style didn't matter so much for the tests. I'll it now that you mention it. Should I make it varargs on the type?

jacobsi:
lgtm

PR #133, opened 2016-06-29
Evaluate timers for results in balboa-http

cramertj:
Fixed a bug where the timers' `toString` was being returned as the response body in some balboa-http endpoints.


jacobsi:
lgtm

PR #135, opened 2016-07-01
EN-7683 Add timing to all balboa-http endpoints that interact with cassandra

cramertj:
I also did some cleanup to remove EntitiesRest and make the type system do some more checking for me (changed Any -> ActionResult in servlet methods).


balboa-http/src/main/scala/com/socrata/balboa/server/MainServlet.scala
```diff
@@ -37,9 +43,32 @@ class MainServlet extends ScalatraServlet
...
-    response.result
+  get("/entities*")(getEntities)
+  def getEntities: ActionResult = {
+    contentType = json
+
+    val predicate: (String => Boolean) =
+      params.get("filter").map(Pattern.compile)
+        .map { pattern => (str: String) => pattern.matcher(str).matches }.getOrElse(_ => true)
+
+    timer("entities queries")({
```
jacobsi: What about `entities-get` as the name? The truth is I'm not entirely sure everywhere this ends up downstream, but this name should (if I understand things correctly) be the name of the metric. I think it might be nice if it was key like.

cramertj: That's true. How about `getEntity`?

cramertj: This is actually the ~forbidden~ API for fetching a list of all entities that have existing metrics. Results look like the following:

```
[
  "testMetricsEntity-603eecde-1769-4463-b585-1ce1f84be5bd",
  "testMetricsEntity-55f7ca73-fcf3-4fe4-ba44-62ead5db8219",
  "testMetricsEntity-2f605cd2-3fe1-4203-b11a-951a9ce0b49b"
]
```
So, it is actually getting entities, or at least their names.

jacobsi:
lgtm

PR #138, opened 2016-07-05
Switch from scala-logging-slf4j to updated version scala-logging

cramertj:
This is a blocker for getting Gatling load testing integrated into balboa, as Gatling depends on scala-logging which is incorrectly resolved when scala-logging-slf4j is included.


jacobsi:
lgtm

PR #140, opened 2016-07-06
Disable balboa-agent integration tests on Jenkins.

jacobsi:
The integration tests for balboa-agent include a clean up step that
attempts to remove the ActiveMQ queue used by the tests. When there is
no ActiveMQ broker available, that steps hangs infinitely because the
ActiveMQ client libraries by default do infinite retry on a connection.

https://socrata.atlassian.net/browse/EN-7819 is to address this issue and fully enable the integration tests.


raganw:

PR #142, opened 2016-07-06
cramertj/EN-7684 Add load tests to balboa-http

cramertj:


balboa-http/src/it/scala/com/socrata/balboa/server/load/PostLoadTest.scala
```diff
@@ -0,0 +1,44 @@
...
+import com.socrata.balboa.server.{Config, EntityJSON, MetricJSON, ServiceUtils}
+import io.gatling.core.Predef._
+import io.gatling.core.structure.ScenarioBuilder
+import io.gatling.http.Predef._
+import org.json4s.{DefaultFormats, Extraction, Formats}
+import org.json4s.jackson.JsonMethods.{pretty, render}
+
+import scala.concurrent.duration._
+
+class PostLoadTest extends Simulation {
```
jacobsi: A quick description here of the intent of these tests. Just something like testing the balboa-http write path by adding <kinds of> metrics.

Since Gatlin is not widely used in our company, the documentation bar is a little higher. People will not necessarily arrive here with a rough general knowledge in place.

balboa-http/src/it/scala/com/socrata/balboa/server/load/QueryLoadTest.scala
```diff
@@ -0,0 +1,34 @@
+package com.socrata.balboa.server.load
+
+import com.socrata.balboa.server.Config
+import io.gatling.core.Predef._
+import io.gatling.core.structure.ScenarioBuilder
+import io.gatling.http.Predef._
+
+import scala.concurrent.duration._
+
```
jacobsi: Same kind of comment here is helpful.

jacobsi:
Can you add a `Load Testing` section to the README that covers this? A quick blub about the tool selected (Gatlin), what is covered (balboa-http, the calls that are covered, kinds of scenarios), and an example of running the tool, including how to run in pointing at a different server.

balboa-http/README.md
```diff
@@ -45,3 +45,6 @@ sbt balboa-http/test
 sbt balboa-http/run "-Dbalboa.config=balboa-agent/src/main/resources/config/config.properties"
 sbt balboa-http/it:test
 
+### Load Tests
+sbt gatling-it:test
```
jacobsi: How do I point this at a different server? I think it should work like this:

    SERVICE_HOST=127.0.0.1 SERVICE_PORT=2012 sbt gatling-it:test

but give it a try.

cramertj: It takes the address from the config, same as the normal tests and integration tests.

cramertj: So yes, setting those env variables works.

jacobsi: Add that to the docs. It's a pretty normal scenario for someone to show up here and want to run this against some arbitrary endpoint. As a user, I would just want to see an example to copy and paste, I wouldn't want to dive into the code to figure out where the setting is used, where it came from and how to set it. If `sbt gatling-it:test --help` could be made to do the right thing, that would be best, but a decent example in the README is second best.

jacobsi:
lgtm

PR #143, opened 2016-07-08
cramertj/request logging

cramertj:


balboa-http/src/it/scala/com/socrata/balboa/server/load/ReplayFileLoadTest.scala
```diff
@@ -0,0 +1,8 @@
+package com.socrata.balboa.server.load
+
+/**
+  * Created by taylor.cramer on 7/7/16.
+  */
```
jacobsi: There is no need for comments like this. Git log has this information.

jacobsi:
lgtm

balboa-http/src/it/scala/com/socrata/balboa/server/load/ReplayFileLoadTest.scala
cramertj: Yeah, that was auto-gen'd by Intellij. I rm'd that file-- it was meant for another branch but got auto-added.

PR #150, opened 2016-07-13
Enable metrics reporting in balboa-http.

jacobsi:
Internally collected application metrics (counters and timers relating
to the functioning of balboa-http itself) are published into graphite
using JmxReporter, which sends the metrics to the collectd instance in
the Docker container (which handlers transferring them on to the final
destination.)


raganw:

PR #151, opened 2016-07-15
Stop setting the JMX port.

jacobsi:
The necessary JMX settings to correctly integrate with collectd metrics
collections are inherited from the socrata/runit-java8 image in the form
of JAVA_TOOL_OPTIONS. This change removes the balboa image specific
overrides.


TranquilMarmot:
:shipit:

jacobsi:
lgtm

PR #152, opened 2016-07-15
Change service metric names.

jacobsi:
The names of timing metrics used for reporting the behavior of the
balboa-http service were preserved in their original format for
backwards compatibility. However, after spending much time
troubleshooting and trying to make metrics publishing actually work, I
can only conclude it never worked, and there is nothing to be backwards
compatible with, so I'm changing the metrics names to something more
useful.


TranquilMarmot:
👍 :shipit:

PR #153, opened 2016-07-26
Add a default uncaught exception handler

rjmac:
Best guess for the balboa-agent hangs are uncaught exceptions terminating
threads.  Rather than messing about with finally blocks, this just adds
a global handler to dump the logs.


znep:
🐑 once tests pass.

PR #154, opened 2016-07-28
Allow BalboaJms log level to be set from the commandline

cramertj:


raganw:

PR #155, opened 2016-07-28
Fix setting log level from command line

raganw:


cramertj:
Squirrel for you sir: :shipit:

PR #157, opened 2016-07-29
Fix the incorrect session handling of the metricPublisher

cramertj:
Also did a refactor so that we aren't depending so much on the word "queue", which is an implementation detail rather than what the thing does, which is publish metrics.


raganw:
Looks good to me :shipit:

PR #159, opened 2016-08-15
Allow configuring log level in container

raganw:


housepage:
:sheep: 🍧

cramertj:
@raganw You tested that setting that env value overrides the log level set [here](https://github.com/socrata-platform/balboa/blob/master/balboa-http/src/main/resources/config/config.properties) / [here](https://github.com/socrata-platform/balboa/blob/master/balboa-http/src/main/resources/log4j.properties)? There's no logic for handling that parameter in [main](https://github.com/socrata-platform/balboa/blob/master/balboa-http/src/main/scala/com/socrata/balboa/server/Main.scala) from what I can tell.

PR #160, opened 2016-08-16
Hack commandline log-level setting into balboa-http

cramertj:
This is a stopgap solution to resolve EN-9166 that should
be replaced by the upcoming work to change the way balboa
is configured (possibly using typesafe config).


raganw:
Fingers crossed! :shipit:

PR #163, opened 2016-09-07
Cramertj/add missing amq conf to balagent

cramertj:


raganw:
Looks good to me :shipit:

PR #164, opened 2016-09-09
Healthcheck returns 503 when unhealthy

cramertj:


raganw:
Looks good to me :shipit:

PR #178, opened 2017-03-17
EN-14710: Also close the file channel

raganw:


raganw:
@housepage gave the thumbs up.

PR #181, opened 2017-05-17
Increase metaspace to more reasonable default.

housepage:
Add metaspace args for balboa-service-jms.


raganw:
Fix it, Felix! Fix it!

PR #186, opened 2017-10-17
EN-18847: Split those metrics!

TranquilMarmot:
Some refactoring:
- `ActiveMQReceiver` -> `ConsumerPool`
- `Consumer` -> pulled from inner `Listener` class from `ActiveMQReceiver`
- `TransportLogger` -> pulled out of inner class from `ActiveMQReceiver`

New `Producer` class will split a message into chunks and re-add it to the queue. Next time the message is pulled off of the queue, it will be small enough to consume & persist to cassandra.


socrata-pull-request-updater:
This PR should be reviewed by: @socrata/missioncontrol

housepage:
It sounds like we have exhausted the timebox on an integration test for this. Since we have a manual repro of the issue, :shipit:

PR #188, opened 2019-03-19
EN-31158: Don't compile docs for Java 7...

TranquilMarmot:


project/BuildSettings.scala
```diff
@@ -27,10 +27,6 @@ object BuildSettings {
       else
         Some("releases"  at nexus + "libs-release-local")
     },
-    // For the time being, leave this as targeting Java 7 bytecode so that
-    // consumers of balboa libraries do not have to update their JVM to get the
-    // newest balboa libraries.
-    javacOptions ++= Seq("-source", "1.7", "-target", "1.7", "-Xlint"),
```
housepage: It feels like we should change to a different target bytecode rather than getting rid of this altogether.

TranquilMarmot: `-target` isn't a valid argument anymore, which is why I'm making this PR

PR #192, opened 2019-12-05
EN-36732: Avoid logging stacktrace line-by-line

rantler:
This is to help with Sumo being able to ingest these log lines without
putting each line of the stacktrace in its own log line.


rantler:
I was finally able to build and run balboa locally and point core to it. However, I'm not sure if the logging format is correct. It logs exceptions in the same manner as Core does, which I presume does not suffer from the SumoLogic parsing problem, so maybe this Just Works™ now.

PiSoup:
Seems like it has a reasonable chance to work and is extremely low risk so YOLO.

PR #196, opened 2020-10-26
EN-42390: Make new config to optionally stop writes from Balboa to Cassandra

bnguyen12:
* New optional ENV variable to stop writes from Balboa (specifically `balboa-service-jms`) to Cassandra
* Typo fixes

### DIFF
--- a/balboa-service-jms/src/main/java/com/socrata/balboa/jms/activemq/Consumer.java
+++ b/balboa-service-jms/src/main/java/com/socrata/balboa/jms/activemq/Consumer.java
@@ -18,11 +18,13 @@ public class Consumer implements MessageListener {
     private DataStore dataStore;
     private Producer producer;
     private int metricCountLimit;
+    private boolean stopWrites;
 
-    public Consumer(ActiveMQSession session, String queueName, DataStore ds, int metricCountLimit) throws NamingException, JMSException {
+    public Consumer(ActiveMQSession session, String queueName, DataStore ds, int metricCountLimit, boolean stopWrites) throws NamingException, JMSException {
         this.dataStore = ds;
         this.session = session;
         this.metricCountLimit = metricCountLimit;
+        this.stopWrites = stopWrites;
 
         Queue queue = session.createQueue(queueName);
         consumer = (ActiveMQMessageConsumer) session.createConsumer(queue);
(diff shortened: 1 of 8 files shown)


balboa-service-jms/src/main/java/com/socrata/balboa/jms/activemq/Consumer.java
```diff
@@ -37,28 +39,34 @@ public void onMessage(Message payload) {
         long start = System.currentTimeMillis();
         String messageText = null;
         try {
-            TextMessage text = (TextMessage) payload;
-            messageText = text.getText();
+            // Before we even consider chunking a large message, check if we even need to write to send out the message
+            if (stopWrites) {
+                session.commit();
+                log.debug("Config was set to stop writes from Balboa to Cassandra. Ignoring message");
```
bnguyen12, line 45: This was the only thing added to this file. The git diff makes this look ugly tho

balboa-service-jms/src/main/java/com/socrata/balboa/jms/activemq/Consumer.java
```diff
@@ -37,28 +39,34 @@ public void onMessage(Message payload) {
         long start = System.currentTimeMillis();
         String messageText = null;
         try {
-            TextMessage text = (TextMessage) payload;
-            messageText = text.getText();
+            // Before we consider chunking a large message, check if we even need to write to send out the message
+            if (stopWrites) {
+                session.commit();
```
bnguyen12, line 44: is this even necessary? if no changes are being made to Cassandra then a commit isn't necessary? or are commits something you should regularly do, even during a NOOP?

progman32: The session is related to ActiveMQ, not Cassandra. I believe we should commit here, else (afaik) the message won't be acknowledged. https://activemq.apache.org/how-do-transactions-work

progman32 (approved):

balboa-service-jms/src/main/java/com/socrata/balboa/jms/activemq/Consumer.java
bnguyen12: Ah ok good, I was wondering how messages were acked/nacked here

